PR #1, opened 2016-02-17
Allow stacking of overlapping items.

meikoudras:
This changes a few things to achieve this result.
screen shot 2016-02-17 at 09 43 10

Overlapping items are stacked on top of each other when they overlap and the areas are resized to match the content. 

This is achieved with **stackItems** property. To make this work I had to refactor all the dimension calculations before the components are mounted and according to the lineHeight set in the properties. All the dimension and stacking calculations are now done in the utils module.

I also changed the main script entry to the new modules folder I created. It is in package.json, the entry looks like this: `"main": "modules/index.js",`

You can generate an updated modules folder with npm run module. The reason for that is to avoid having to run Babel for both your own code and the libraries you are using. I discovered that with certain setups of Babel (older versions) the include libraries are not compiled correctly. By using the compiled version by default including the library in a webpack setup becomes much more simpler. (I recommend to do that with the scss files as well. Currently the library has a dependency of libsass and your webpack setup must be ready to parse scss files. By using the compiled version by default it would only become a development dependency.)

I also made it possible to change the header height with the two new properties **headerLabelHeight** and **headerLabelGroupHeight**. These are currently set to the default value of **lineHeight**. 

I also added a property called: **useResizeHandle**. What this does is add a small handle on the right side of the timeline item that you can use to resize the item instead of relying on the edges of the item. This has the appropriate css styling to infer that this will resize the item. This still needs a bit of work since it doesn't work very well with very small item sizes. I suspect it has to do with the interact.js library.

I also added new clickhandler logic to the timeline component which will allow for more complex (react components) content in the timeline items. Currently only text content was working correctly with interactions.

I also added the property **traditionalZoom** which means that the scroll button/trackpad scroll will zoom in and out, instead of moving around in the timeline. You can also now click and drag the timeline around. I also added the possibility to zoom in with alt + scrollwheel since ctrl + scrollwheel is a default accessibility zoom on mac os x which overrides all other controls. This should probably configurable by property.

Please review the changes and make suggestions what should be improved or changed.


lnagel:
@mariusandra - what do you think?

mariusandra:
Hey!

This looks amazing! I played around with it and do find it very very nice and useful! :+1: 

I do have a few comments and questions though. Nothing too serious, I hope. :)

First about the changes:

**1) Stacked items**
Great feature!

Could you update the default demo (`npm start` and `http://localhost:8080`) to also use stacked items? This seems like a great feature to advertise! 

If I were to nitpick, I can bring out three comments:

a) The gaps between the items are pretty big (see the line "Beverly"):
screen shot 2016-02-18 at 10 50 57

The first line for "Beverly" looks visually closer to the line from "Jordan". How easy would it be to bring all the lines for "Beverly" closer together? 

b) While dragging, could the line remember that it was once 2 items high, so that this would not happen:

ezgif-3636477864

c) Could it be possible to fix the item on the line, it's being dragged on? Like in this video, it's not intuitive that the item is sometimes on line #1, sometimes on line #2.

ezgif-2426629793

It would be really cool if when dragging it over the 3rd line, it stays on the 3rd line. Then when letting go, it can snap to wherever makes sense in the row.

**2) Modules**

Great for the modules system. I wasn't aware that this is how it should be done. Feel free to convert the styles as well to this system. The less dependencies in the host projects, the better.

**3) Header height**

Great!

**4) Resize handles**

Great!

I have removed resizing when the item is below a certain width. Perhaps if these handles don't work well for small items, they can also be removed?

**5) Traditional zoom**

With this enabled, is it possible to scroll up and down on the page at all, if it's e.g. a list of 100 rows? Or then you need to use the arrow keys and a scroll bar?

With the possibility to drag to scroll (great!) and alt+scroll to zoom, plus the dates on the headers to zoom in/out, I can only find this scroll-to-zoom to be confusing. Since zooming in is a computationally expensive operation as well, I would avoid it being the default behaviour that can be triggered automatically.

Could you also explain for which users/browsers/computers this zoom is a better option? Any point in creating an option to automatically detect them?

...

Besides those comment on the functionality, I have a few more about the code:

**6) Code style**

I'd like to enforce consistency in the coding style. Up until now, all code for this repository was written using the [JavaScript Standard Style](http://standardjs.com/). Apart from various common sense suggestions, the biggest thing it requires is omitting semicolons. I find code without semicolons *feels* nicer, and since the transpiler/minifier anyway adds/removes them based on context, there's no real reason to use them.

While I used the `linter-js-standard` atom package myself to ensure this, I have now added a more tranditional `.eslintrc` file, and the required packages to `package.json` on `master`.

Please run `npm run lint` to see what the style complains about, and then `npm run lintfix` to fix most of the issues automatically. Feel free to activate eslint for your editor, if you haven't already.

In `master` I see a few issues that `eslint` complains about, for which `standard` was silent, like useless constructors. I left them in for now, in order to avoid potential merge conflicts.

**7) lodash**

I see you have added lodash as a dependency. I would like to avoid that, or if not possible, include it in a different form.

Lodash is an amazing utility, but it's just massive, at 62KB minified. It makes the size of the library balloon from ~40kb to ~100kb. There are 3 options I see:

a) Remove lodash entirely and rewrite the `_.keyBy`, `_.groupBy` and `forEach` functions using ES6. `_.forEach` can be replaced with a 2-liner, `_.keyBy` with a 3-liner and... not sure for `_.groupBy`.

b) Instead of including the enetire `lodash` module, include `lodash.groupby`, `lodash.keyby` and `lodash.foreach`. This will already save a lot of bytes.

c) Put `lodash` as a peer dependency. However I'd discourage that, as there is no guarantee the "host project" uses or wants to use lodash.

I'd prefer option A. :)

**8) speed**

This package was originally architected in such a way, as to limit expensive computations to the minimum. That's why there is a hidden "canvas" behind the scenes that is 3 times the width of the screen. This way, when scrolling the screen 50% to the left or 50% to the right, all the items stay in place, and we can use the browser's hardware accelerated scrolling. This is a lot faster compared to an approach, where there is no hidden canvas and all the elements get their {position: 'absolute', left: `${x}px`} moved on every frame. React's DOM diffing is great, but it is just too slow for such a workload on every scroll event. Now the "big update" only happens when you scroll the screen more than 50%. 

This is why there are two different times that are kept in the component's state: visibleTimeStart/-End and canvasTimeStart. 

I noticed that in `Timeline#render` you call the `calculateDimensions` and `stack` functions on every `render()`. My non-scientific testing shows, that this adds about 10-20ms to the time it takes to render every frame: from ~50-60ms to ~60-80ms.

Is it possible to move those operations to a callback, that's triggered when the `canvasTimeStart` state variable is changed? It should speed things up a bit.

**9) functions under render()**

Under `TimeLine#render`, there are many functions being called, like `{this.horizontalLines(...)}`, `{this.sidebar(...)}`, etc. They call the specific react components and pass along parameters. I see some of these parameters come from inline `this.state.x` and `this.props.x` calls, some from parameters being passed to the functions.

Is there any rule, that says when something is passed as `width={this.state.width}` or sent through the function's parameters and passed as `canvasWidth={canvasWidth}`? I'm not disagreeing with this, just looking for consistency :).

---

Phew, that was longer than I expected...

All in all I think this is great work! The component is already usable. With these final changes, I'm ready to merge the PR.

meikoudras:
1) Agree with all the points. Will try to find a solution.

4) I think it currently does disable the resize handles when the element is too small. But I would have to check.

5)  This was done only so that the zooming works like it works in  [vis.js.](https://github.com/almende/vis/) I do not feel strongly about it being like that. You can scroll when you hover over the list of names on left. I think the current scrolling/zooming system is better for usability than the traditional zoom, which is why I left it disabled by default. The ALT + Scroll zooming I find is necessary for people who have accessibility options enabled they are not able to zoom at all.

6) Yep, will fix linting issues.

7) Agree. Lodash should be removed as a dependency.

8) I have currently not looked at optimising the new placement algorithms, but I will take a look at it and try to implement the changes you suggested.

9) Honestly I think the function calls are unnecessary even since they all return React components. They should just be react components in the render method. Since it was originally written like that I didn't change it. Also when it comes to using it state and prop variables then the better style is (as I have read): if you use the variable in calculations and don't just pass it then use destructuring to get the value and then pass the destructured value. Otherwise use the full variable e.g `width={this.state.width}`

mariusandra:
Hey,

9) I wasn't sure of the right way to do things when I made those functions. My goal has always been readability and avoiding ~100 lines of JSX tags. It was a lot messier before. That said, if you can find a way to merge the JSX tags into render() while retaining or improving code clarity, go ahead :).

As for destructuring. This is debatable. I find it really helps code clarity if everything a function subscribes to (all public state/props variables it accesses) is clearly defined at the top, even if the variables are just passed through. This draws a nice dependency graph and reduces the likelyhood of strange bugs, as most are caught by the linter.

malindamiller:
@meikoudras This feature is very useful, however I am looking to have one group at a time stack.  So for instance, the default of stackItems would be set to false, but then a button would allow that specific group to stack or unstack.  Do you have any pointers for getting started on this?

PR #5, opened 2016-03-18
Add doubleclick handler on canvas

meikoudras:
Allow doubleclick listener on canvas. When you double click on the canvas the click handler is given the time position where the click happens and the group. Ideally it allows creating of new items to the click position.


mariusandra:
0.7.3 released!

signalwerk:
is it a good idea to have `onCanvasClick(groupId, time, e)` but `onCanvasDoubleClick(time, groupId)`? Would it not be better to have `onCanvasDoubleClick(groupId, time, e)`? see: https://github.com/namespace-ee/react-calendar-timeline/pull/52

PR #9, opened 2016-04-12
[Core] Updated dependencies. Fixed warning in Header.jsx

xmityaz:
Updated dependencies, including react 15.0.*
At least it works correctly for me. Probably would be better to test deeper, as I could miss something ('npm run lint' in terminal doesn't work for me either with new and with old dependencies)

Fixed warning in Header.jsx:


ochervak:
Cool! Many thanks for your PR

lnagel:
This package should remain usable both in projects with react 0.14.* or react 15.0.*.

ochervak:
what about accepting ? ;)

madisvain:
We'll review this shortly. Sorry the delay.

mariusandra:
Hey, I checked and everything seems to work. 0.7.6 pushed out.

Issue #6, opened 2016-03-30
When prefetch new data, error in console appears, but timeline works fine


ochervak:
I use timeline component in next way:
```
<Timeline groups={this.getResidents()}
                          items={this.getActivityTimeline()}
                          itemHeightRatio={1}
                          defaultTimeStart={this.props.startTime}
                          defaultTimeEnd={this.props.endTime}
                          canMove={false}
                          dragSnap={60 * 60 * 1000}
                          lineHeight={40}
                          canChangeGroup={false}
                          onBoundsChange={this.onCanvasBoundsChange}
                          visibleTimeStart={this.props.startTime.valueOf()}
                          visibleTimeStop={this.props.endTime.valueOf()} />
```

Inside `this.onCanvasBoundsChange` i use ajax call to server, and new items from `this.getActivityTimeline()` pushed to your beautiful component.

However in Chrome debugger i expect `es6.promise.js:119 Unhandled promise rejection TypeError: Cannot read property 'dimensions' of undefined(…)`

And it refs to transpiled line `dimensions: sortedDimensionItems[(0, _utils._get)(item, itemIdKey)].dimensions,` which is in `visibleItems.map(function (item) {` inside `render`.

Can you explain this error, or maybe i using your tool in some wrong way?

Thanks in advance

mariusandra:
Hey!

Does the fix by @xmityaz, which is now in v0.7.8 solve the issue for you?

ochervak:
yes, it solved! thanks

jerryslaw:
Guys I still have this problem, same case. My component is re-rendering every time, when new context will come (setting new state from context, parsing data for items, and passing items from new state to Timeline).

PR #10, opened 2016-04-20
[Timeline] Update dimensionItems when [items] or [groups] props are changing.

xmityaz:
#6 
I've faced the same issue when was trying to add dynamically new items to the timeline.
This made a trick for me. Hopefully, this will also fix the issue above


mariusandra:
Hey, thank you!!! I checked the code and it looks good to me!

Verison 0.7.8 pushed out with some small code style tweaks (`npm run lintfix`).

PR #16, opened 2016-04-27
support for ie-9

weslley39:


mariusandra:
Hey,

Sorry to hear you still have to support IE9. However there are a few issues with this PR.

1) I assume the package `babel-preset-es2015-loose` should be added to `package.json`?
2) I'm not sure using loose mode is a good idea. From [this page](http://www.2ality.com/2015/12/babel6-loose-mode.html): "Normally, it is recommended to not use loose mode.".

Since I don't know any better, I want to be cautious of this and avoid problems down the line due to support for a 5-year old browser.

If, however, you do need to use this with loose mode, I recommend the following.

1. Set up `es2015-loose` in your project like you normally would.
2. Redefine the `react-calendar-timeline` alias in your webpack conf to point directly at the source JS, like so: 

```
  resolve: {
    alias: {
      "react-calendar-timeline": path.join(__dirname, 'node_modules', 'react-calendar-timeline', 'src', 'index.js'),
    }
  }
```

In this case you will not use the "normal mode" transpiled code (from `react-calendar-timeline/modules/index.js`), but use the ES6 code directly, and transpile it using your own babel settings through your project.

PR #22, opened 2016-05-07
[Item] Added property for handling contextMenu event on item

xmityaz:
In our project we need to overwrite the context menu on an item. 
Probably it can be useful for somebody else :)


mariusandra:
Thank you for the PR! v0.8.1 released!

PR #23, opened 2016-05-11
[Item][Events] Send original event object to handler function arguments for mouse events

xmityaz:
Sometimes it can be useful to have original event object inside handler method arguments. 
For example to get coordinates of the mouse cursor on rightClick / doubleClick and open custom context menu at the appropriate place.

In this PR I added this just for simple mouse events. Probably in future, it will make sense to do the same for interact.js events (for such handlers like onItemMove, onItemResize).


mariusandra:
Thank you! 👍

Issue #24, opened 2016-05-15
Item gets timezone offset on dragging and resizing if timeline has dragSnap property.


xmityaz:
Reproducible link: http://codepen.io/xmityaz/pen/wGZRoq
To reproduce the issue try to slightly move one of the items (Item 2 and Item 3 are more obvious). It will start dragging process but at first will move/resize the item on timezone offset value (3 hours for Kiev) and just after on `dragSnap` value (1 day).

The issue occurs when the `dragSnap` property is set to Timeline and it isn't multiplier of timeZone offset value.
I guess the problem in `dragTimeSnap` function in Item.jsx (line 63). 
I've tried to fix it in the way bellow (changed dragTimeSnap function):
```
dragTimeSnap (dragTime) {
    if (this.props.dragSnap) {
      const offset = moment().utcOffset() * 60 * 1000;
      return Math.round(dragTime / this.props.dragSnap) * this.props.dragSnap - offset % this.props.dragSnap;
    } else {
      return dragTime
    }
  }
```

But it solves the issue just for moving an item. And introduces new issue for resizing:
For [this expamle](http://codepen.io/xmityaz/pen/wGZRoq) if you will try to resize Item 1 it will reset item time to 12 AM (instead of 6 AM in this case).

Could you please help to solve this issue if you'll have some time?
Any way, if I'll find a solution earlier I'll be back with PR :)

mariusandra:
Thanks for the PR! I just released 0.8.2 with it!

PR #30, opened 2016-05-22
[Item][dragSnap] Fixed item move and resize behavior when dragSnap property set

xmityaz:
Fix to #24 issue.
In `resizeTimeSnap` method we consider `endTime` which is indivisible part of `itemTimeEnd` to `dragSnap` property. So the item length will always increase or decrease equally on `dragSnap` property and will save the correct time.
Also added `considerOffset` parameter to `dragTimeSnap` method to fix item drag behavior.

I've tested this with different time zones and `dragSnap` values and it works for me.

### DIFF
--- a/src/lib/items/Item.jsx
+++ b/src/lib/items/Item.jsx
@@ -60,11 +60,23 @@ export default class Item extends Component {
     return (props.canvasTimeEnd - props.canvasTimeStart) / props.canvasWidth
   }
 
-  dragTimeSnap (dragTime) {
-    if (this.props.dragSnap) {
-      return Math.round(dragTime / this.props.dragSnap) * this.props.dragSnap
+  dragTimeSnap (dragTime, considerOffset) {
+    const { dragSnap } = this.props;
+    if (dragSnap) {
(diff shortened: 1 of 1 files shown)


mariusandra:
Merged!

signalwerk:
Hey @xmityaz I have also a kind of a time-offset problem. Do you have an opinion about that; https://github.com/namespace-ee/react-calendar-timeline/issues/56
Should we fix that? For me it's not suitable like it is right now but what's your experience with it?

xmityaz:
Hi @signalwerk 
I think #56 is more bug then feature.
It probably should be fixed in the same place with this PR. At some point we should take difference between timezones (for dayligth saving time they should be different), multiply it on miliseconds in hour (1000*60*60) and reduce this value from dragTime and resizeTime.
You can try to create new PR for this issue, but I'm not sure it'll be merged soon.

signalwerk:
@xmityaz sorry for the late response and thank you for your answer. I'm going to put it on my todo list. It's right now not prioritized but at one point I'll work on it. I'll report back here.

PR #36, opened 2016-07-08
Fixed memory leak when attach\deattach touch[start,move,end] listeners

ochervak:
Hi, as far as i know, .bind(this) !== .bind(this), so your handlers will never be de-attached. That cause memory leak in my application.

This PR eliminate this issue.

BR


mariusandra:
Thank you! Merged!

PR #66, opened 2017-01-31
Fixed header width and Header label weekday support

meikoudras:
When using the fixed header option in a container that has horizontal padding the header labels container overflows out of the calendar area. When using the fixed width of the calendar area as the header container width (instead of 100%) this issue does not appear. 

Also added the same day of the week classes as vertical columns for styling the header labels as well.


mariusandra:
Thanks! Merged and it will be out with 0.13.0. :)

PR #71, opened 2017-02-13
ADD: Option for multiple selected Items

meengit:
If you pass the prop `selected=['id1','id2']` to the timeline component the default behaviour of selecting only one item at a time is overwritten. From then on you have to manage the selected items yourself within the `onItemSelect` handler and pass the new id's to the selected props again. This enables you to have multiple items selected. This change is not breaking the existing behaviour because if `selected` is not set nothing changes for the user.


mariusandra:
Thanks! Merged and out with 0.13.0

Issue #39, opened 2016-07-24
Container width in %


flyalex:
Hi,

I am dividing my screen in to 2 parts. First contains timeline, second - other content. 
Second part has width param, first uses rest of space.

If second part has fixed size in pixels, timeline works great even when changing size of the browser.
But if second has width set in %, part of timeline disappear  accidentally (everything besides groups names).

Any Idea, how to solve it?
Thanks!

ochervak:
+1 that i have same problem.
i fixed it with calling `.resize()` method when layout changes. it it dirty hack, but it works in my case

BTW i use mobx

```
// Subscribe for layout changes
        this.dispose = observe(
            // Subscribe for my updates
            ui.filters.state[types.OVERVIEW_FILTER],
            (change) => {
                // If change is to open/close filter (which require re-painting)
                if (change.name === 'opened') {
                    // If we hide filter -> save initial width of timeline
                    if (!change.newValue) {
                        change.object.memorizedScrollWidth = this.refs.timeline.refs.scrollComponent.offsetWidth;
                        setTimeout(() => this.refs.timeline.resize(), 0);
                    // If we show filter -> restore saved initial width for timeline
                    } else {
                        this.refs.timeline.refs.scrollComponent.style.width = `${change.object.memorizedScrollWidth}px`;
                        setTimeout(() => this.refs.timeline.resize(), 0);
                    }
                    ui.showDebug('Trigger resize of overview');
                }
            });
    }
```

mafzst:
Any news here?

mariusandra:
Hey, while I couldn't reproduce the bug as you described, there is now support to detect changes to the width of the container itself. Here's [a demo](http://namespace.ee/react-calendar-timeline-docs/#/elementResize) where it works ([source](https://github.com/namespace-ee/react-calendar-timeline/blob/master/demo/app/demo-element-resize/index.js)). 

Closing this issue. Feel free to reopen if there are still issues.

PR #75, opened 2017-03-08
Calculate width when we receive sidebar width property

jmerriweather:
This fixes issue #39 
also fixes the FIXME issue "currently when the component creates a scroll the scrollbar is not used in the initial width calculation, resizing fixes this"


jmerriweather:
I've found some issues with my fix

mariusandra:
Thank you! It's finally merged and will be out with 0.13.0

jmerriweather:
My pleasure! 👍

signalwerk:
😻

PR #80, opened 2017-03-30
Added an option to show one more sidebar on the right

goooseman:
new feature

I have added an option to show one more sidebar on the right.

This is completely optional, by default the library works absolutely the same.

But, if you need the second sidebar, you can do the following:

Add two props to the `<Timeline />` component:
```
      rightSidebarWidth={150}
      rightSidebarContent={<p>Second filter</p>}
```

And add `right_sidebar` prop to the groups objects:
```
{
  id: 1,
  title: 'group 1',
  right_sidebar: 'additional info about group 1'
}
```

-------------------

I have added new functionality, updated README, updated version number in the package.json, added info to the CHANGELOG, built `modules` and `build`, linted code, so this PR is completely ready to merge.

Looking forward for your feedback.

Best wishes,
Alexander 😉


madisvain:
This looks totally cool.

What would be some of the use cases that you see this feature used for?

goooseman:
You can pass not only a string, but the component, so you can render some additional information for each row and add a second filter. For example, like this:
screen shot 2017-03-31 at 9 59 05

signalwerk:
@goooseman your example screen looks rad! Wow! Do you have that code by any chance online? Or is it private?

goooseman:
@signalwerk this is a picture of the design, which I'm implementing right now. It is not done yet)

goooseman:
@signalwerk It will be published approximately in two months

madisvain:
@goooseman @signalwerk ok I see what the use case is there. We'll be looking into making this available with react-calendar-timeline.

mariusandra:
This looks rad! Merged and out with 0.13.0

PR #86, opened 2017-05-02
Update README.md

nicocrm:
Add documentation for visibleTimeEnd, visibleTimeStart, canvasTimeEnd and canvasTimeStart parameters (tripped me up because the defaultTimeStart / End are moment objects but these are timestamps)


signalwerk:
@nicocrm thank you for your PR! I'm working right now on bringing different PR's to a new version. But yours is now already merged since there is no code affected.

PR #87, opened 2017-05-03
Avoid updating updateDimensions right after updateScrollCanvas

nicocrm:
Because setState runs at the end of updateScrollCanvas and applies the update asynchronously, updateDimensions computes using stale state.  When visibleTimeStart and visibleTimeEnd are specified this causes the items to be drawn in the wrong place occasionally when scrolling.  This update fixes it.  Since updateScrollCanvas already updates the item dimensions the call to updateDimensions is redundant anyway.


signalwerk:
hey @nicocrm thank you for your pull request. looks good. I'll try to push it to master with version 0.11.3 beginning next week.

mariusandra:
Thanks for the PR! I have merged it and it'll be out with 0.12.1!

mariusandra:
actually... it'll be out with 0.13.0 because of #103 :)

PR #92, opened 2017-05-07
FIX: Typo in readme. no coffee-beans here

signalwerk:
only changes on the readme therefor no version bump


signalwerk:
since nobody complained... I'm going to merge it.

PR #94, opened 2017-05-08
Add support for element-resize-detector

Ziller321:
Now uses element-resize-detector for resize detection.
This can be changed back to 'window' resize eventListener with property `resizeDetectorMethod ="window"`.

Benefit of this is that when something resizes Timeline component for example side scroll bar, we could detect this change and call resize for the component.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "parser": "babel-eslint"
   },
   "dependencies": {
+    "element-resize-detector": "^1.1.12",
     "interact.js": "^1.2.6",
     "react-pure-render": "^1.0.2"
   },
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -9,6 +9,7 @@ import Header from './layout/Header'
 import VerticalLines from './lines/VerticalLines'
 import HorizontalLines from './lines/HorizontalLines'
 import TodayLine from './lines/TodayLine'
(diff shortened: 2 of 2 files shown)


signalwerk:
@Ziller321 thank you for your PR! looking good! I just started to maintain the project a 2-3 weeks ago and I'm working on the hanging Pull Requests. I'll take care of yours as well. Please give me a bit time. Thank you.

mariusandra:
Hey, this looks good! My only concern: it adds another [18KB minified](https://github.com/namespace-ee/react-calendar-timeline/pull/94) dependency making this component balloon from ~[http://namespace.ee/react-calendar-timeline/demo.bundle.js](60kb) to ~80kb.

So if possible, I'd like to somehow make this an optional dependency that you can add on if needed. 

Something like

```
import Timeline from 'react-calendar-timeline'
import elementResizeDetector from 'react-calendar-timeline/lib/resize-detector'

<Timeline resizeDetector={elementResizeDetector} ... />
```

so that webpack and others could skip it if not used.

Does anyone have any thoughts on this?

signalwerk:
Sounds right to me. Over time we should somehow find a "API" how to extend the component without bloating it. But props is probably the best idea right now.

mariusandra:
Hey, I converted this to a pluggable version and merged it into 0.14.0. Thanks! :)

PR #95, opened 2017-05-10
Update package.json

nicocrm:
Update "module" command to work with GNU sed... tiny change but the command was not working on my linux box :)


signalwerk:
macOS eats both. so looking good to me. thanks a lot!
what do you think about using node-sass? maybe you have an opinion about that; https://github.com/namespace-ee/react-calendar-timeline/pull/97

PR #97, opened 2017-05-11
change for module from scss to node-sass 

signalwerk:
To keep it consistent with the webpack build it would be better to also take node-sass


nicocrm:
Yes it seems that would be more convenient too, right now you have to install sass as a separate step.

signalwerk:
Thank you for your expertise and the review! I'm going to merge it therefore.

PR #103, opened 2017-05-26
Custom item renderer

nicocrm:
Add groupRenderer and itemRenderer props to allow using custom component in the timeline


mariusandra:
Hey, I reviewed the code and it looks good! I have merged the changes and they will be out with 0.13.0!

PR #106, opened 2017-06-12
FIX: renders with empty groups-array

signalwerk:
If the Timeline had no grups;

```jsx
<Timeline
  groups={[]}
  items={items}
  defaultTimeStart={moment('1995-12-25').add(-12, 'hour')}
  defaultTimeEnd={moment('1995-12-25').add(12, 'hour')}
/>
```

the component threw a error. I handled this case in `stackItems` because this gets executed already in the constructor. Someone a better idea where to handle it?

### DIFF
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -853,6 +853,18 @@ export default class ReactCalendarTimeline extends Component {
   }
 
   stackItems (items, groups, canvasTimeStart, visibleTimeStart, visibleTimeEnd, width) {
+
+    // if there are no groups return an empty array of dimensions
+    if(groups.length === 0) {
+      return {
+        dimensionItems: [],
+        height: 0,
+        groupHeights: 0,
+        groupTops: 0,
+      };
+    }
+
+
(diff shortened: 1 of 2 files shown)


mariusandra:
Hey, seems good enough for me. Thanks! :)

signalwerk:
thx!

signalwerk:
@mariusandra Ups! Sorry for not linting it. Thx for eef5955726e45adce1cb5582e82e6851c1fb06ba

mariusandra:
@signalwerk np! 

I don't think this was a problem at all. It's good to lint every once in a while, but at this stage I wouldn't require it from contributors before accepting PRs. I'll be anal and do it anyway before a new version if I see semicolons :).

signalwerk:
Well at one point we should have travis do the lining and run the tests to get a feedback to the PR for the contributors. But it's nice you brought de deploy script up and running! great

MertHaddad:
How about rendering with empty items ? the timeline is not showing if my items array is empty.

PR #107, opened 2017-06-13
FIX: if items with no corresponding group

signalwerk:
```js
let groups = [
  {
    id: 1,
    title: 'Title'
  }
]
let items = [
  {
    ...
    group: -1, // here is the problem!
    title: 'Title'
  }
]
```


mariusandra:
Thanks! :)

signalwerk:
🎉

PR #108, opened 2017-06-13
ADD: example for itemRenderer and groupRenderer

signalwerk:
is there a reason why in the `index.js` the additional props are wrapped in `itemProps`? It's one way right? But could also be passed in as unwrapped additional attribute? At least it works for me.

```jsx
let items = [
  {
    id: 1,
    group: 1,
    title: 'Title',
    tip: 'additional information',  // <<<< like here
    ...
  }
]
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,8 +340,8 @@ Here is an example that limits the timeline to only show dates starting 6 months
 
 ```js
 // this limits the timeline to -6 months ... +6 months
-var minTime = moment().add(-6, 'months').valueOf()
-var maxTime = moment().add(6, 'months').valueOf()
+const minTime = moment().add(-6, 'months').valueOf()
+const maxTime = moment().add(6, 'months').valueOf()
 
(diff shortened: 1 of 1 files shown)


mariusandra:
Hey! Thank you for the doc updates!

The `itemProps` came in with PR #58 and they're passed as attributes to the `<div />` created by `<Item />`, like so: https://github.com/namespace-ee/react-calendar-timeline/blob/05be8ec1026ab5c0b261a11f509d67c23118300e/src/lib/items/Item.js#L461
```jsx
    return (
      <div {...this.props.item.itemProps}
           key={this.itemId}
           ref='item'
           ...
       />
    )
```

I guess this can be used to pass additional tags and handlers to the item `<div />`, for example ARIA tags, rich snippets, random event handlers (`onDoubleClick`?), etc.

Not sure. I don't use this. @sjchmiela - what do you use `itemProps` for?

signalwerk:
Thx!

PR #111, opened 2017-06-17
check group order for undefined in getGroupedItems

nicocrm:
The items from the first group don't get rendered properly, because they have an order of 0 and the getGroupedItems method only does a falsy comparison


mariusandra:
Thanks! 0.14.5 published with the fix!

signalwerk:
with test 😻 thx a lot!

PR #112, opened 2017-06-19
Update CHANGELOG.md to include 0.14.2

rkait:
Added 0.14.2 with the changes for prop-types.

Also some of the compare links are broken, so I added it for direct commits. Just fyi! Thanks for the changes @mariusandra.


signalwerk:
Thanks a lot for your PR! Great! 🎉

Issue #85, opened 2017-05-02
DragSnap value is used when computing the item width, sometimes leading to wrong width calculations


gabriel-deliu:
It seems that the dragSnap value is used when computing the item width even if items can't be resized.
`let w = Math.max(itemEnd - itemStart, dragSnap) //react-calendar-timeline/src/lib/utils.js line 186`
This will cause items smaller than the dragSnap value (default 15 min) to be displayed incorrectly as dragSnap long.

Setting a low value to dragSnap is a workaround.

signalwerk:
@gabriel-deliu do you have by any chance an example for me?

gabriel-deliu:
Having dragSnap set to 15 minutes (default), create a 10 minute event. 
Expected: The event should be displayed as taking 10 minutes.
Actual: The event is displayed as taking 15 minutes.

Workaround: Set dragSnap to 10 minutes or lower.

signalwerk:
Oh! I worked today on `calculateDimensions` (https://github.com/namespace-ee/react-calendar-timeline/pull/113). I completely get you now. I also made a [code example](https://www.webpackbin.com/bins/-KnBvRcj2FzvwL8QbT01) to illustrate it. The two Items in Group B are 10min and 15min. I can't work this week on the component anymore but I'll fix that soon. should be easy.

mcMickJuice:
I believe this has been fixed as of 0.17.2 . If not, I'll reopen this.

PR #113, opened 2017-06-21
Refactoring of calculateDimensions

signalwerk:
I'm working on a function to have multiple Items drag at the same time. But for that I don't wanna mess with the `calculateDimensions`. Therefore I had a look at it and worked on some simplification and made it testable. 

_The following things were done in this PR:_
* no need to filter by valid dimension after `calculateDimensions`. Only add  if dimension is valid (`reduce instead` of `map` & `filter`)
* everything that is not pure calculation is taken out of the function so we don't need the keys in the `calculateDimensions`
* no round trips where we pass values in just go give it back then (like for `stack`)
* added tests

Beside the `lineHeight` I didn't do anything to the general handling. The `lineHeight` was just adjusted because we don't need to save it in each calculation. 

Over all it should be faster and hopefully more clear.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,6 @@ bower_components
 .idea
 .sass-cache
 modules
-lib
+/lib
 build
 gh-pages
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -971,31 +971,45 @@ export default class ReactCalendarTimeline extends Component {
     const visibleItems = getVisibleItems(items, canvasTimeStart, canvasTimeEnd, keys)
     const groupOrders = getGroupOrders(groups, keys)
 
-    let dimensionItems = visibleItems.map(item => {
-      return {
-        id: _get(item, keys.itemIdKey),
-        dimensions: calculateDimensions({
-          item,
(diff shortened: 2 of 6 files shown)


mariusandra:
Wow, nice! I'll have a look at it tomorrow morning latest :)

mariusandra:
For future reference:
I found this code: `dimension.stack = !item.isOverlay ? ...` which seems to be an undocumented feature. Nothing in the README references `isOverlay`. This should be added one day.

signalwerk:
@mariusandra I completely understand what you say about [Timeline.js#L997-L1001](https://github.com/namespace-ee/react-calendar-timeline/blob/master/src/lib/Timeline.js#L997-L1001). It feels a bit dirty. But I have in general not the best feeling about passing something like `dimensions.top = null;` but I didn't wanna change too much for the moment. 

Since there is no functional change it should be a patch-version I guess. I'll not work this week anymore on the component so no need to wait for anything. 

but https://github.com/namespace-ee/react-calendar-timeline/issues/85 will be fixed next week.

I'm still a bit a mess with the npm. But I thought I understood now how I can push from my repo right to the npm. Sorry I skipped again a version. But that should not happen anymore. Sorry.

PR #117, opened 2017-06-23
ADD: transform-runtime because Safari doesn't support Object.entries

signalwerk:
Right now the Demo didn't work anymore on Safari for Mac because Object.entries is right now not supported (right now only in demo, but probably coming more and more). Maybe it's anyway better to have the runtime included to have a better browser support. I don't like it too much because the native functions are way cooler but I guess it's also nice to have it running on more browsers.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,6 +1,7 @@
 {
   "presets": ["es2015", "stage-0", "react"],
   "plugins": [
-    "react-hot-loader/babel"
+    "react-hot-loader/babel",
+    "transform-runtime"
   ]
 }
--- a/package.json
+++ b/package.json
@@ -65,6 +65,7 @@
     "babel-eslint": "^7.1.1",
     "babel-loader": "~6.2.4",
     "babel-plugin-add-module-exports": "^0.2.1",
+    "babel-plugin-transform-runtime": "^6.23.0",
     "babel-preset-es2015": "^6.6.0",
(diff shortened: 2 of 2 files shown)


mariusandra:
Hi, the Object.entries call in the demo was actually useless, so I removed it: https://github.com/namespace-ee/react-calendar-timeline/commit/9b6aea93a5be8b617adb17d9b5722866c68665d9

Some comments:
1) Yes, I'll start doing PRs as well. It's about time! This was the last regular commit :)
2) Is `transform-runtime` the right tool for this? From what I read in [their docs](https://babeljs.io/docs/plugins/transform-runtime/), it polyfills entire objects (`Promise`, `Map`, etc) and gathers redundant babel calls. On the top of the page they say: *"NOTE: Instance methods such as "foobar".includes("foo") will not work since that would require modification of existing built-ins (Use babel-polyfill for that)."*. Adding `Object.entries` is doing just that.

signalwerk:
1. don't worry. since I'm sometimes not that fast with PRs you can feel free to do as you like.
2. I actually also don't like the runtime too much... OK let's agree that we don't use it and just avoid stuff like `Object.entries`. Thank you for taking it out with https://github.com/namespace-ee/react-calendar-timeline/commit/9b6aea93a5be8b617adb17d9b5722866c68665d9

Issue #74, opened 2017-02-27
How can I change the background color of cell?


gyarasu:
Hello.
I try to use your great component.

I have our calendar data which include milestones (e.g. our regular release days).
I want to change the background color of cells based on our calendar data.
It seems that weekend cells color are changed by default.

![figure](https://cloud.githubusercontent.com/assets/4948588/23347281/6463dd7c-fce4-11e6-989a-16eaf91a8fbb.png)

Is it possible?

Thanks!

signalwerk:
@gyarasu there is a feature called `ranges` developed by @meengit 
with it you can mark some ranges in the background of the timeline component. right now it works but it's not 100% ready for a pull request. but you can already use it. with ranges you can do something similar what you are asking for. Here you finde the repo: https://github.com/meengit/react-calendar-timeline/tree/feature/range#ranges  (branch: feature/range)

Here you see two ranges:
![ranges](https://cloud.githubusercontent.com/assets/992878/23355474/1faf8554-fcd6-11e6-8eee-d7bf871ea8d3.png)

gyarasu:
Thanks for your response.

I successfully installed #ranges branch into my project
and I confirmed that I can add the property.

I have a question.
How should I set a color for ranges?

I added the ranges property like below.(other properties are omitted this time)
```
<Timeline 
    ranges={{
          id: 2,
          start_time: moment().add(-1, 'day'),
          end_time: moment().add(4, 'day'),
          className: 'weekend'
     }
/>
```

Next, I prepared css.
```
.weekend {
  background-color: #99cc00
}
```

Unfortunately, I can't see it seems to be something wrong.

signalwerk:
Hmmm... the current branch has other property names. That will be fixed before the final pull request comes. Here my Ranges definition:

```JSX
<Timeline 
    ranges={[{
          id: 2,
          start: moment().add(-1, 'day').toDate(),
          end: moment().add(4, 'day').toDate(),
          className: 'weekend'
     }]}
/>
```

result:
![demo](https://cloud.githubusercontent.com/assets/992878/23485649/a0e25fc2-fedc-11e6-96b0-ab4b1e9f3f8b.png)

fopsdev:
Just an additional question.
I have a use case where i need to grey out the background as ranges are doing it.
But the ranges are per employee (Valid Worktime is defined differently for every employee)
Is it possible?
Thanks for any enlightment

signalwerk:
[...]

PR #122, opened 2017-06-28
Plugins

### LINKED ISSUE #77: Is it possible to link items?
I'd like to show several items in a group as linked. 

I have a repeating event that I'd like to be able to style to show those events are linked.

Awesome module, thanks!

mariusandra:
Hey! I experimented with adding an abstract (background?) `painter` component that could be used to draw arbitrary things on the timeline: cell and row backgrounds, lines connecting items, webgl accelerated particles or videos of Kiefer Sutherland jumping into Christmas trees... In theory you could render anything without having to modify the timeline itself.

Here's a demo that uses it to display different background colors for rows:

screen shot 2017-06-28 at 10 11 21

It could be extended to add the `ranges` feature you mentioned in #74 or to fix #77. 

Basically the custom `painter` component would get a lot of internals from the timeline as props (`groupHeights`, `dimensionItems`, etc), and use this to do whatever is needed. 

Any thoughts on this approach?


signalwerk:
@mariusandra hey this is really cool! Wow! I reviewed this week some code of @meengit and he planed to send a PR for ranges. but with that method in place that's not necessary anymore. Instead he should have it as a separate npm as a "plugin" to load. 
So we should probably have two options for this prop. You can pass in a react-component or an array of react-components. So you can load multiple plug-ins

I would suggest the following API:

## untouched
* `canvasTimeStart`, `canvasTimeEnd` - start and end of the scrolling canvas in Unix timestamps
* `canvasWidth` - width of the scrolling canvas in pixels
* `visibleTimeStart`, `visibleTimeEnd` - start and end of the currently visible area
* `groups`, `items`, `keys` - groups, items and keys as passed to the timeline
* `dimensionItems` - an array of objects `{ id, dimensions: {...} }` describing positions of all the items
* `height`, `headerHeight` - height of the entire calendar (includes `headerHeight`) or just the header
* `timeSteps` - steps for displaying time

## new
* `dimensionGroups`- an array of objects `{ id, dimensions: {...} }` describing positions of all the groups
* `selected` - an array of Item ID of selected items. ((I swapped internally all to array handling > PR is coming next week. So there is no big difference between passed selected or internal selection handling. Therefore I would only give that one props to make it more uniform))

## drop
* ~~`lineCount` - number of groups~~ this could be done with `dimensionGroups.length`no?
* ~~`groupHeights`, `groupTops` - arrays of heights and tops for the groups~~ we have `dimensionGroups` ((this line is 2x in the readme.md)
* ~~`selectedItem`, `selected` - the selected item or controlled selected items array~~ 
 
I'm right now on an other project till mid next week so I can't invest too much time right now. But I try to keep up with the comments.

mariusandra:
Hey, thanks for the review!

Indeed after this, the `ranges` PR could and should be converted to a plugin. Your idea of allowing many plugins to be added at the same time sounds great. In fact, I'm thinking of redesigning the API to allow for this flexibility. My proposal is to allow something like this:

```jsx
<Timeline {...props}>
  <Ranges ranges={[/* 3pm to 5pm every day */, /* another range */]} />
  <RainbowBackground />
  <SomeOtherPlugin />
</Timeline>
```

All of these plugins would get cloned and passed these extra props. Similar to how I did it for [pigeon-maps](https://github.com/mariusandra/pigeon-maps).

With `v0.13.0` I deprecated passing the content for the left sidebar as children to the component, so there should be nothing in the way of this.

This might even pave the way for a fully composable timeline, where you could just do 

```jsx
import Timeline, { HorizontalLines, VerticalLines, TodayLine, Sidebar } from 'react-calendar-timeline'
import Ranges from 'react-calendar-timeline-ranges'

<Timeline {...timelineProps}>
  <Sidebar alignment='left' />
  <HorizontalLines />
  <VerticalLines />
  <TodayLine />
  <Ranges {...rangeProps} />
</Timeline>
```

... but that's to be thought of later. For sure before reaching 1.0 :).

For the other comments

* I'll wait for your PR on the `selectedItem` vs `selected` changes.
* Agreed about `lineCount` not being useful
* For merging `groupHeights` and `groupTops` into `dimensionGroups`, it makes some sense. However I would also like to avoid calls to create new objects just to pass to `painter`. I guess I could refactor the stacking functions to already return and use this combined object.

mariusandra:
Hey... and I did those API changes!

* The children of the component now get extra props... and thus I removed the `painter` prop.
* I merged `selectedIndex` and `selected` to just `selected`
* I removed lineCount

[Here's some code](https://github.com/mariusandra/react-calendar-timeline/blob/painter/demo/app/demo-plugins/index.js#L111)

I still didn't do anything in regards to `dimensionGroups`

README.md
```diff
@@ -424,8 +421,23 @@ import containerResizeDetector from 'react-calendar-timeline/lib/resize-detector
...
+
+The children gets passed the following props:
+
+* `canvasTimeStart`, `canvasTimeEnd` - start and end of the scrolling canvas in Unix timestamps
+* `canvasWidth` - width of the scrolling canvas in pixels
+* `visibleTimeStart`, `visibleTimeEnd` - start and end of the currently visible area
+* `groups`, `items`, `keys` - groups, items and keys as passed to the timeline
+* `height`, `headerHeight` - height of the entire calendar (includes `headerHeight`) or just the header
+* `groupHeights`, `groupTops` - arrays of heights and tops for the groups
+* `groupHeights`, `groupTops` - arrays of heights and tops for the groups
```
signalwerk: this line is double

signalwerk:
I fucking like that new idea of yours;

```jsx
<Timeline {...props}>
  <Ranges ranges={[/* 3pm to 5pm every day */, /* another range */]} />
  <RainbowBackground />
  <SomeOtherPlugin />
</Timeline>
```

this is awesome. 

I would suggest to keep working on the  API. Stuff lie `dimensionGroups` is a bit work to rewrite our internal use. but it would also be much clearer. 

But yeah. Just merge it and we keep an eye on it. I send you tomorrow the PR of my [multiselect/multidrag feature]/(https://github.com/signalwerk/react-calendar-timeline/tree/feature/multiselect) and next week I try to adopt @meengit s ranges to a plugin. then I have a better view if the API fits for simple cases

mariusandra:
Cool! Thanks for the review! I'll add a notice in the README that the API for plugins will still change, so be prepared.

And I propose to move in the direction of unifying the internal and external APIs.... so we could also get rid of all the [pornographic methods](https://github.com/namespace-ee/react-calendar-timeline/blob/master/src/lib/Timeline.js#L1184-L1201) in Timeline.js that just render components (`this.items`, `this.verticalLines`, etc).

mariusandra:
I merged the code and will do a release with this (and the new header format) in the evening (CET). One more thing: I still need to write tests for this. That will also have to wait for the evening :).

signalwerk:
@mariusandra completely agree with your words! [eat your own dog food is always good](https://en.wikipedia.org/wiki/Eating_your_own_dog_food)

Don't hurry for the release. We are not in a rush. :)

mariusandra:
Well, it's merged and 0.15 is out! It think it's a good base for further improvements.

I also added [a simple test](https://github.com/namespace-ee/react-calendar-timeline/blob/master/src/lib/__tests__/index.js#L105-L139) to at least make sure the children get the right props. Once we settle on what those props should actually contain, we can extend it further.

Let the flood of new plugins begin! 🌊 😁

Issue #83, opened 2017-04-13
Touch scroll do not work


Shadowman4205:
When enable inspect in chrome in mobile scroll do not respond at all? any idea?

eksavoy:
You can fix this issue with properties :
fixedHeader="fixed"

Shadowman4205:
it is workingnow but it mess up the header.
Cause i have got  a navbar on top and in the middle i have got timeline.

Shadowman4205:
@mariusandra Maybe any suggestion? how make touch scroll work without making headear fixed?

mariusandra:
@Shadowman4205 Hey, a fix is on its way. Once PR #125 is merged you will no longer need `fixedHeader="fixed"` for the touch scroll. I'll update you when that happens!

Shadowman4205:
thank you very much!

Shadowman4205:
Sorry for inconvience is the PR #125 merged?

signalwerk:
@Shadowman4205 sorry for the waiting. @mariusandra asked me to review the PR. I'm going to do that within the next 12h.

Shadowman4205:
@signalwerk 
lol do not tell sorry. Thank you very much!

mariusandra:
@Shadowman4205 Hey, I have just released `v0.15`. Please try it out and let us know if it fixed the problems you were having.

You should no longer need to specify `fixedHeader='fixed'` for the touch scroll to work. Also, the default header mode has changed to `"sticky"`. If that's too much, you may always reset it to `"none"`.

Closing this issue now :)

PR #125, opened 2017-06-28
Flexible header

mariusandra:
*(This PR is based on the "painter" branch, as I did some zIndex reordering there... so if/when  #122 gets merged, the changed files for this PR will be much cleaner...)*

Hey, 

When investigating issue #83 I tried to clean up issues people have been having with the header and its positions. Here are the changes I did:

- I completely removed the broken `fixedHeader='absolute'` and replaced it with `fixedHeader='flexible'`, which I believe should be the new default and works like this: https://youtu.be/Z2hEpSW-Yj8
- I noticed that we had two different scroll/zoom modes. First if `fixedHeader` was `fixed`, the component took control of the scrolling and zooming behaviours. It waited for scroll/touch events and moved the scroll canvas accordingly. If `fixedHeader` was anything else, we let the user's browser manage the scrolling/zooming/pinching of the scroll canvas and then we reacted to it. While this made the scrolling faster, rendering elements on the canvas after the scroll was much slower (see comparison [here](https://youtu.be/WkBwaKzV8gY?t=21s) around second 21). So I removed the special case for the "non-fixed header", which also caused issues such as #83. Now we always control the scroll and everything should be visually smoother.
- I added a demo that shows the flexible scroll mode.

Any comments?

I guess once this and the `painter` branch get merged, it's time to do a new minor release. :)


signalwerk:
@mariusandra hey thank you very much for this PR. It contains now some painter code and the header. I focuse here only on the header.

* There is a merge conflict righ now. Easy one. But if you don't feel like here is my branch with it. Just pull from it. [signalwerk:flexible-header](https://github.com/signalwerk/react-calendar-timeline/tree/flexible-header)
* for the prop `fixedHeader` I would not call the new option `flexible`. I would call it `sticky` because it's more or less like the [sticky positioning](* https://drafts.csswg.org/css-position/#sticky-pos ) in the new css proposal. [Example in Chrome](https://codepen.io/tutsplus/pen/gpggXb)

other than thta all good for me. thank you for your work

mariusandra:
Great, thanks for this! I'll merge it later today and release a new minor version. And good idea with renaming `flexible` to `sticky`. It does make more sense!

PR #131, opened 2017-07-06
Feature/multiselect

signalwerk:
This implements a multiple select and multiple drag option. I changed the whole logic away from `dragTime` to `dragTimeDelta`. But when it comes to callbacks or `calculateDimensions` I kept the old logic, so not everything changes. If you could give it a run if I caught everything I would be thankful. 

multiselect

### DIFF
--- a/demo/index.html
+++ b/demo/index.html
@@ -5,7 +5,6 @@
     <meta content='width=device-width, initial-scale=1' name='viewport'>
     <link href='https://fonts.googleapis.com/css?family=Lato' rel='stylesheet' type='text/css'>
     <title>react-calendar-timeline</title>
-    <link rel="stylesheet" type="text/css" href="demo.css">
   </head>
   <body>
     <div id="root"></div>
--- a/src/lib/items/Item.js
+++ b/src/lib/items/Item.js
@@ -137,7 +137,7 @@ export default class Item extends Component {
     if (this.props.moveResizeValidator) {
       dragTimeDelta = this.props.moveResizeValidator('move', item, this.itemTimeStart + dragTimeDelta) - this.itemTimeStart
     } else {
-      dragTimeDelta = Math.round(dragTimeDelta / dragSnap) * dragSnap
+      dragTimeDelta = (Math.round((this.itemTimeStart + dragTimeDelta) / dragSnap) * dragSnap) - this.itemTimeStart
     }
 
     if (onDrag) {
--- a/src/lib/items/Item.js
+++ b/src/lib/items/Item.js
@@ -28,6 +28,7 @@ export default class Item extends Component {
     //
     // onSelect: React.PropTypes.func,
     // onDrag: React.PropTypes.func,
+    // onDragStart: React.PropTypes.func,
     // onDrop: React.PropTypes.func,
     // onResizing: React.PropTypes.func,
     // onResized: React.PropTypes.func,
@@ -48,8 +49,7 @@ export default class Item extends Component {
       interactMounted: false,
 
       dragging: null,
-      dragStart: null,
-      preDragPosition: null,
+      dragStartPos: null, // holding x,y position when draging
       dragTime: null,
       dragGroupDelta: null,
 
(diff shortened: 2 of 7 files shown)


signalwerk:
BTW: I have right now not really a clue how to test something like this...
My current test-implementation: no tests  :see_no_evil:

demo/index.html
```diff
@@ -5,7 +5,6 @@
     <meta content='width=device-width, initial-scale=1' name='viewport'>
     <link href='https://fonts.googleapis.com/css?family=Lato' rel='stylesheet' type='text/css'>
     <title>react-calendar-timeline</title>
-    <link rel="stylesheet" type="text/css" href="demo.css">
```
mariusandra: While this gives annoying errors in development, it's actually needed in [production](http://namespace.ee/react-calendar-timeline-docs/#/). Not sure how to make both worlds happy...

mariusandra:
Very cool! 

I had a really quick skim through the code just now (hence the one comment) and will go though in depth in the next days. Also no idea how to test it...

What happens or should happen if you resize when you have multiple items selected?

src/lib/items/Item.js
```diff
@@ -118,11 +118,67 @@ export default class Item extends Component {
...
+    }
+
+    this.setState({
+      dragTimeDelta,
+      dragGroupDelta
+    })
+  }
+
+  dragend (e) {
+    console.log('item dragend')
```
mariusandra: this probably shouldn't be here :)

src/lib/items/Items.js
```diff
@@ -101,6 +93,34 @@ export default class Items extends Component {
...
+  itemDrop = (itemId, dragTimeDelta, oldGroupOrder, dragGroupDelta) => {
+    const { itemIdKey, itemGroupKey, itemTimeStartKey } = this.props.keys
+
+    const groupOrders = getGroupOrders(this.props.groups, this.props.keys)
+
+    this.props.selectedItems.forEach((selectedItemId) => {
+      const item = this.props.items.find(itemObj => _get(itemObj, itemIdKey) === selectedItemId)
+      let order = groupOrders[_get(item, itemGroupKey)]
+
+      console.log('items itemDrop')
```
mariusandra: this probably shouldn't be here :)

mariusandra:
Hey, I checked the code and played around a bit. Overall it looks very nice. The refactoring from `selectedItem` -> `selected` is very welcome!

The one big comment I have is that something changed with the `dragSnap` property when there's just one item selected.

This is the main demo with the live `0.15` version: 

old-version

This is with the multiselect branch:

new-version

The FPS of the gifs is not too great, but in short while previously the items snapped to 15min slots, now there's no snapping at all. Plus when dragging the height of the group shrinks a bit, to indicate that the group is empty. 

This snapping remains in the multiselect demo, but works a bit differently than before: Until now a `dragSnap` of 15min made it that the post-drag start of the items was in multiples of 15min. So dragging an element that starts at 18:32 resulted in it landing on 18:15, 18:30 or 18:45. In the multiselect demo the element will land on 18:17, 18:32 or 18:47.

I think it should work like before, at least when dragging just one element. I know there's some issue (#85) with this dragSnap and item lengths, so it might be time to agree on some behaviour and document it, for both dragging and resizing.

Once this gets fixed, I'll be happy to merge :). 👍

signalwerk:
thank you for the review! I didn't notice the wrong handling of the `dragSnap`. I completely agree with the logic of the component so far. I think these should be the rules:

* `dragSnap` should be relative to the start time of an item
* if multiples selected the item dragged snaps and the other just get the offset. <br> Example:<br>Item A starts 12:15, Item B starts 12:20. <br>After the drag of both at the same time with `dragSnap` 15 min:<br>Item A starts 12:30, Item B starts 12:35
* resizing while having multiple selected is possible but effects only the one selected. If people wanna prohibit that behaviour or do the resize for all items they can do that in their own logic with `moveResizeValidator` and `onItemResize` 

I'll work on Thursday on the component again.

Thanks again for your time to have a closer look at my code!

((btw we should probably rename `onItemResize` at one time. that sounds like its a callback during resizing. it should better be `onItemResized`))

signalwerk:
I corrected the two problems you mentioned. The moving of items now snaps correct and the problem with the height was related to the groupdelta. I kept the resizing logic like before. But we should probably also change here the logic that it snaps like moving items.

mariusandra:
Hey! Sorry for the radio silence so far. I got caught up in a very busy period...

I'll review the changes next Monday or Wednesday.

One quick comment without looking at the code though. Going with this: "`dragSnap` should be relative to the start time of an item" would change the current behaviour I believe.

IIRC, the current behaviour (snapping to :00, :15, :30, :45 even if the item is at :32) was like this so that the component would also clean up bad data. E.g. the :32 was entered through a timepicker that was too loose... or perhaps you cloned some old item and it kept the time... and then you could at least fix it by dragging. I have to check how this change would affect our app here.

However, perhaps this is something that really depends on the usecase and a flag like `snapTo="absolute|relative" would help?

mariusandra:
Hey, I finally managed to have another look at this! 
One thing that I found is that in the multiselect demo, if I select more than one item and drag them down so that one of the elements goes outside the calendar, [things break](https://www.youtube.com/watch?v=PaAh_BS0X44&feature=youtu.be). Is this a bug with the demo or with the implementation? I believe this should be somehow limited in the component

I also noticed that the times now [snap to the "grid"](https://www.youtube.com/watch?v=ifUSLuwyF-k&feature=youtu.be), which is consistent with the current implementation. 👍 

Once this dragging-past-the-end error is fixed I don't see anything in the way of a merge :)

vasdee:
see https://github.com/namespace-ee/react-calendar-timeline/pull/314 for a new pull-request that is this pull, refactored with the 0.16 branch.

mcMickJuice:
Closing in favor of https://github.com/namespace-ee/react-calendar-timeline/pull/314

PR #161, opened 2017-08-23
Left and right sidebar unification

valgussev:
Hi Marius, here you'll find the left and right sidebar unification with an option to add more than one sidebar to both sides. I'm relatively new to React, so any comments are very welcome.

In a nutshell I've added a `sidebars` array prop to the Timeline component which consists of:
id, sidebarWidth, sidebarContent, sidebarKey and isRightSidebar. So now you can add as many sidebars as you wish on both sides.

Thanks in advance.

### DIFF
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -19,7 +19,6 @@ import { getMinUnit, getNextUnit, getParentPosition, _get, _length, stack, nosta
 export const defaultKeys = {
   groupIdKey: 'id',
   groupTitleKey: 'title',
-  groupRightTitleKey: 'rightTitle',
   itemIdKey: 'id',
   itemTitleKey: 'title',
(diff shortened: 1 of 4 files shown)


mcMickJuice:
Hi @valgussev thanks for this PR!  Would you mind opening an issue for this feature and explaining the use case?  Not saying we won't merge it but before we implement a feature, it's nice to know the  motivation and use case for said feature.

Also, this feature changes the props passed into Timeline as well as changes how the user defines a sidebar. Can you update the README to capture this change?  We'll also want to have backwards compatibility here so we'll still need to accommodate users passing in the individual sidebar props.

mcMickJuice:
I'm going to close this PR.  We can always reopen but I don't see the value of adding this functionality into the core library.  If you'd like, please open an issue and if it gets enough attention we can revisit this.

PR #199, opened 2017-12-28
Add scrollableContainer prop to allow using sticky header behavior with non-window scrollable elements

unutranyholas:


src/lib/Timeline.js
```diff
@@ -498,7 +500,10 @@ export default class ReactCalendarTimeline extends Component {
   }
 
   componentWillReceiveProps (nextProps) {
-    const { visibleTimeStart, visibleTimeEnd, items, groups, sidebarWidth } = nextProps
+    const { visibleTimeStart, visibleTimeEnd, items, groups, sidebarWidth, scrollableContainer } = nextProps
+    if (scrollableContainer !== this.props.scrollableContainer) {
+      throw new Error('scrollableContainer update is not implemented')
```
mcMickJuice: What do you think about setting `this.scrollableContainer` in the constructor from props and never update `props.scrollableContainer`? We can be explicit in the README that this prop is a onetime setting?

I just don't like the idea of throwing and borking a user's app if they change a prop. I do think we should have a `console.warn` to tell the user that they're doing something wrong.

src/lib/Timeline.js
```diff
@@ -81,6 +81,7 @@ export default class ReactCalendarTimeline extends Component {
     minResizeWidth: PropTypes.number,
     fixedHeader: PropTypes.oneOf(['fixed', 'sticky', 'none']),
     stickyOffset: PropTypes.number,
+    scrollableContainer: PropTypes.object.isRequired,
```
mcMickJuice: Make this propType not required since we have a default value in defaultProps.

mcMickJuice:
@unutranyholas  @aodinok im not able to reproduce this issue with the current library version (0.15.6).  Here's my attempt at reproducing:

https://codesandbox.io/s/mqrm74ykkp

If you could reproduce this issue in this sandbox, please do so I can move forward with this PR

aodinok:
I've changed the code in https://codesandbox.io/s/2jqmnr67jy to reproduce the issue. You can also double-click after scrolling to see that the wrong group is passed to the handler as an argument because of the wrong offsetTop value.

mcMickJuice:
@aodinok have you pulled down this branch and confirm these changes work?  I'm trying to pass the scrolling container using ref to the `scrollableContainer` prop but there is a timing issue here:

* refs are resolved before `componentDidMount` is called on component
* `componentDidMount` bubbles up (timeline `cDM` is called before parent `cDM`). 
* we register the scroll listener to scrollable container in cDM of timeline.

I think i can just look at `componentWillReceiveProps` and update listener to the new container thats passed in? Maybe?

aodinok:
Yes I did pull off this changes and tested them in my environment. But I'm using timeline component in angular.js project, so I'm passing DOM node directly by document.getElementById. I don't have access to computer today, will test this changes tomorrow in cases with refs, and get back to you.

aodinok:
Tested this today with React component. You were right @mcMickJuice, it doesn't work with refs as expected. I managed to make it work by removing exception throw and replacing it with the code that removes event listeners from the old scrollableContainer and then attaches it to the new one.

`  if (scrollableContainer !== this.props.scrollableContainer) {
      this.props.scrollableContainer.removeEventListener('scroll', this.scrollEventListener)
      scrollableContainer.addEventListener('scroll', this.scrollEventListener)
    }`

Here is sticky demo with scrollableContainer set to div:

`  <div style={{padding: 20, paddingTop: 0, height: '600px', overflow: 'scroll'}} ref={(node) => { this.setState({scrollableContainer: node})}}>
   ....
        <Timeline groups={groups}
                ...
                  scrollableContainer={this.state.scrollableContainer}`

hb3mfr8dlu

mcMickJuice:
lol very nice @aodinok .  That ref callback with setState gives me pause as I remember seeing somewhere that re-renders of the parent component that owns the Timeline could change that node...but with your changes in componentWillReceiveProps will account for that.  I also could be overstating the implications of doing that. 

Overall this is an edge case for using this component and we can document in the README that nuances of using this api have some...weirdness.

src/lib/Timeline.js
unutranyholas: 1. React checks propTypes after it defaults `props` with `defaultProps`
2. We declare `scrollableContainer: window` as default prop
3. If we relax requirement to `non-required object`, then it will be possible to pass `scrollableContainer={null}`, component will fail and this will emit no validation warnings.

Now it works correctly with no false positive react propTypes warnings

unutranyholas:
Updated implementation and documentation, please review

mcMickJuice:
hi @unutranyholas can you please look at https://github.com/namespace-ee/react-calendar-timeline/pull/264 ?  In that PR, I reworked the implementation of sticky header to using css property `position: sticky` instead of the javascript solution (due to performance issues and code complication).  Using `position: sticky` should result in you not having to specify a container since a stickied element will become fixed (i.e. will stick) when it's being scrolled out of viewport.

https://css-tricks.com/position-sticky-2/

With this in mind, I'm not sure this PR is needed.  Can you please confirm though?

mcMickJuice:
Closing as this PR is superseded by changing to position: sticky

PR #205, opened 2018-01-08
use .getAttribute('class') instead of .className

Metallium:
fixes:
```
Uncaught TypeError: e.className.split is not a function
    at y (utils.js:360)
    at t.value (Timeline.js:641)
    at t.D.scrollAreaClick (Timeline.js:1551)
```
e.g. when doing mouseUp on SVGElement child

see https://developer.mozilla.org/en-US/docs/Web/API/Element/className#Notes
> It is better to get/set the className of an element using Element.getAttribute and Element.setAttribute if you are dealing with SVG elements

### DIFF
--- a/src/lib/utils.js
+++ b/src/lib/utils.js
@@ -325,9 +325,11 @@ export function getGroupedItems (items, groupOrders) {
   return arr
 }
 
-export function hasSomeParentTheClass (element, classname) {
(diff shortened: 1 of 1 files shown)


mcMickJuice:
Hi @Metallium thanks for this PR!  Could you please write a test for this?

PR #231, opened 2018-01-19
Fixing issue with wrong offsetTop in state after scroll

aodinok:
Can be merged after this PR: https://github.com/namespace-ee/react-calendar-timeline/pull/199

Now if timeline is nested inside another div with overflow: scroll - it doesn't handle canvasDoubleClick and other click events correctly - because of the wrong offsetTop in state. We should listen to parent scroll and update offsetTop.

### DIFF
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -612,7 +612,7 @@ export default class ReactCalendarTimeline extends Component {
         }
       }
       if (e.deltaY !== 0) {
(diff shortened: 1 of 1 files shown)


mcMickJuice:
Thanks for this. I'll work on getting this and 199 merged

mcMickJuice:
I'm going to close this PR but reopen it if others run into this issue.  I don't think many users are placing the timeline inside a scrollable div and I would encourage them not to do so.

Moreover, the way this timeline currently handles reporting item clicks i believe will change and will not be related to yPosition to determine group

PR #239, opened 2018-02-09
items isSelected to return boolean

Najarana:
item.selected is declared with propTypes to be a boolean. When I supply my own selected prop I get a proptype warning. This is because the isSelected function in the Items component returns a boolean only if there is no selected prop, the problem is that if an array of ids is supplied it returns selected.find() which will return what was found and not the boolean expected by Item.

Using includes instead  will return a boolean value


syakuis:
@mcMickJuice When will this "pool request" be accepted?

mcMickJuice:
Thanks @Najarana . This was an oversight on my part when I enabled propTypes in Item.  Will try to push this release tomorrow morning and including a couple other fixes.

PR #242, opened 2018-02-09
Fix snapping logic for case with timezone offset

aodinok:
We should take utcOffset of dragTime otherwise it will respect current time offset, but it may change depending on date (summer/winter) time


mcMickJuice:
hi @aodinok can you open up an issue where you describe what's wrong/what enhancement you'd like and then link to this PR?

mcMickJuice:
I can't reproduce any bug related to this change.  Closing until an issue is opened with a repro of this bug occurring.

PR #254, opened 2018-03-02
Fix header positioning

mcMickJuice:


mcMickJuice:
fixed issue caused by https://github.com/namespace-ee/react-calendar-timeline/pull/252

Issue #247, opened 2018-02-19
Improve performance of timeline


mcMickJuice:
This timeline performs very well when there are a few number of items passed in, however perf and framerate drops once you reach a non-trivial amount of items (around 50 or so).  This is especially the case when fullUpdate=true.  This thread can be used to list out situations when performance degrades and ideas on how to improve performance.  Obviously reducing the number of items passed to the calendar helps but there should be alternatives/refactors that can be done to support a lot of items by default.

mcMickJuice:
`fullUpdate=true` is generally always desired since we want the item content in each item to be "sticky" when the item is exiting the timeline to the left.  However, it is very expensive as the left value needs to be computed on every render...which occurs on scroll, zoom etc.

Some options/possibilities
- use `position: sticky` on item contents so that computing the left style for items is unnecessary
- only calculate the "dimensions" (i.e. left, top) values of items that are within the viewport (both horizontally and vertically).  This would require a complete refactor of how the timeline works.

mcMickJuice:
maybe use react-virtualized's collection component for rendering of the timeline:

https://github.com/bvaughn/react-virtualized/blob/master/docs/Collection.md

Maybe this is an opt-in

mcMickJuice:
@jabidof what are your use cases when you see performance degradation in the timeline?

jabidof:
@mcMickJuice I'm preparing some directions...
Our main use case is to represent 1000's of items of different activities (possible with different colors, etc.)
See https://i.imgur.com/duWtcpN.png and https://i.imgur.com/MKuYKN8.png

vasdee:
any update on this one? I have a similar requirement as @jabidof over the next few months. If I know the direction I can potentially help out with this one.

AymericG:
react-virtualized would definitely help.

mcMickJuice:
@vasdee sorry for the late response on this. I haven't really had much time to work on this library...plus performance diagnosis and improvement is a long and often frustrating endeavor.

When zoomed out in the timeline and there are a number of items rendered, the timeline becomes unresponsive.  Hell, even when there aren't many items rendered on the timeline, the framerate is pretty poor (<30fps).

Run the project locally and check out the [performance demo](https://github.com/namespace-ee/react-calendar-timeline/tree/develop/demo/app/demo-performance) (fullUpdate is off in this demo)

I've identified a number of areas where this timeline can be approved:

## `fullUpdate` is a performance killer

[...]

PR #264, opened 2018-04-04
Refactor how header is rendered

mcMickJuice:
Part of https://github.com/namespace-ee/react-calendar-timeline/issues/247

This prevents the header elements (days, month units) to rerender and set left and top inline on every scroll (major perf hit)

* moved timeline header and sidebar header components into separate Header container. simplifies header dom slightly
* sync'ed scroll between timeline and header
* removed `fixedHeader` prop in favor of `position:sticky` by default.  fixedHeader and logic associated resulted in some pretty complicated logic


mcMickJuice:
There are some regressions that need to be addressed here:
[] allow for polyfill of position sticky
[] make header scrollable
[] header click on touch devices doesnt work
[x] allow for timeline header to not stick

PR #283, opened 2018-04-22
Minor speed/scrolling optimization

pencilcheck:
After basic testing via `npm start` I can safely say that it performs a lot better under pressure with 500+ groups. The vertical scrolling is a lot smoother if you scroll on top of the canvas, and horizontal scrolling is also much more consistent.

The main idea here is to use `transform` instead of `left/top`.
Also pass in visibleTime{Start/End} for calculating visible items.

Some more ideas in the future:
1. Memoize item components and async update between allowed timeframe for much more steady FPS while scrolling. (e.g. https://medium.com/@planttheidea/memoize-react-components-33377d7ebb6c)
2. Instead of setting scrollTo, perhaps wrap it in requestAnimationFrame so it will be prioritized at a more appropriate times in browser reflow/repaint cycle.
3. https://github.com/joshwnj/react-visibility-sensor for Item component with appropriate polling time period (I tried it locally and it doesn't work since you are using position absolute but however this is still a good idea to dive deeper)

Let me know if there are any questions about this commit. Hope this will be merged soon.

Cheers

### DIFF
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -1317,8 +1317,8 @@ export default class ReactCalendarTimeline extends Component {
 
     const visibleItems = getVisibleItems(
       items,
-      canvasTimeStart,
-      canvasTimeEnd,
+      visibleTimeStart,
+      visibleTimeEnd,
       keys
     )
     const groupOrders = getGroupOrders(groups, keys)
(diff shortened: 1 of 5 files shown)


mcMickJuice:
Pertains to this issue: https://github.com/namespace-ee/react-calendar-timeline/issues/247

pencilcheck:
ok, no problem, I can do that.

And yes I turned off fullUpdate when there are too many items, I made a couple adjustments to find better sweet spot but still not really seeing much improvements though however.

Yes, I thought header update (esp. aligning the dates) is also seems to be the performance hog and slowing down the scrolling a ton at scale.

If you have a better solution cooking up or don't like this change I can remove this from this pull request. Just let me know.

Issue #303, opened 2018-05-07
cut extra renders for performance


Ilaiwi:
Hi, 

I ran a react optimization tool called [why-did-you-update](https://github.com/garbles/why-did-you-update) to find reasons why items render on every little interaction when they shouldn't. This tool runs deep equality for the changing props and state to figure out where props or state differ in reference but not in content. I discovered hundreds of extra render and cutting them is not very hard to fix. Fixes mostly are just changing some components to pure components or prevent `next state` with deep equality to set the upper-level component's state which causes the whole component to rerender.

![pref](https://user-images.githubusercontent.com/10817089/39687568-b301d788-51d7-11e8-90e9-ea60342ee861.gif)

mcMickJuice:
hi @Ilaiwi ill try to review this stuff this week or next. I'm really slammed at work 😄

PR #304, opened 2018-05-07
optimizations

Ilaiwi:
Improvements to eliminate extra components renders. Please refer to #303 for extra information.

An additional improvement would be removing `deepObjectCompare` from `Items` and `Item` and use [immer](https://github.com/mweststrate/immer) when calculating `dimensionItems`, `groupHeights` and `groupTops`

### DIFF
--- a/demo/index.js
+++ b/demo/index.js
@@ -5,6 +5,11 @@ import App from './app'
 
 import './index.html'
 
+if (process.env.NODE_ENV !== 'production') {
+  const {whyDidYouUpdate} = require('why-did-you-update');
+  whyDidYouUpdate(React);
+}
+
 const render = AppToRender => {
   ReactDOM.render(<AppToRender />, document.getElementById('root'))
 }
--- a/demo/index.js
+++ b/demo/index.js
@@ -5,11 +5,6 @@ import App from './app'
 
 import './index.html'
 
-if (process.env.NODE_ENV !== 'production') {
-  const {whyDidYouUpdate} = require('why-did-you-update');
-  whyDidYouUpdate(React);
-}
-
 const render = AppToRender => {
   ReactDOM.render(<AppToRender />, document.getElementById('root'))
 }
--- a/package.json
+++ b/package.json
@@ -125,6 +125,7 @@
     "sass-loader": "~4.0.2",
     "style-loader": "~0.13.0",
     "webpack": "~1.13.3",
-    "webpack-dev-server": "~1.16.2"
+    "webpack-dev-server": "~1.16.2",
+    "why-did-you-update": "^0.1.1"
   }
 }
--- a/package.json
+++ b/package.json
@@ -125,7 +125,6 @@
     "sass-loader": "~4.0.2",
     "style-loader": "~0.13.0",
     "webpack": "~1.13.3",
-    "webpack-dev-server": "~1.16.2",
-    "why-did-you-update": "^0.1.1"
+    "webpack-dev-server": "~1.16.2"
   }
 }
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -24,7 +24,7 @@ import {
   getVisibleItems
 } from './utility/calendar'
 import { getParentPosition, hasSomeParentTheClass } from './utility/dom-helpers'
-import { _get, _length, arraysEqual } from './utility/generic'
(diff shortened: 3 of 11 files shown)


Ilaiwi:
@mcMickJuice I stumbled on a new library that claims to be 3x faster and smaller than `lodash.isEqual` which is [react-fast-compare](https://github.com/FormidableLabs/react-fast-compare). Maybe replacing lodash with this one?

demo/index.js
```diff
@@ -5,6 +5,11 @@ import App from './app'
 
 import './index.html'
 
+if (process.env.NODE_ENV !== 'production') {
+  const {whyDidYouUpdate} = require('why-did-you-update');
+  whyDidYouUpdate(React);
```
mcMickJuice: do you mind removing this block?  This should only be used to profile performance as a means to an end.  I don't want to add a dependency that profiles performance (i.e. it should only be used when performing dev work)

package.json
```diff
@@ -125,6 +125,7 @@
     "sass-loader": "~4.0.2",
     "style-loader": "~0.13.0",
     "webpack": "~1.13.3",
-    "webpack-dev-server": "~1.16.2"
+    "webpack-dev-server": "~1.16.2",
+    "why-did-you-update": "^0.1.1"
```
mcMickJuice: please remove why-did-you-update

src/lib/Timeline.js
```diff
@@ -947,6 +952,32 @@ export default class ReactCalendarTimeline extends Component {
     return label ? <InfoLabel label={label} /> : ''
   }
 
+  renderLeftSidebar = () => {
```
mcMickJuice: Why move these into separate functions? Is this related to optimizations?

src/lib/lines/TodayLine.js
```diff
@@ -12,10 +12,26 @@ export default class TodayLine extends Component {
   }
 
   static defaultProps = {}
+  
+  state = {
+    currentTime: new Date(),
+  }
+
+  componentDidMount(){
+    this.interval = setInterval(()=> {
```
mcMickJuice: Are these changes specific to optimizations?  If not, can you please remove these changes.  I'd like to keep PRs focused in their intent.  

I see where you're going here and I agree with these changes but they should be in a separate PR.

mcMickJuice:
Is the reason you've added deepCompare and arraysEqual because the objects `dimensionItems`, `groupHeights` and `groupTops` are new references every time we render?  If this is the case, we should look at the methods that create these values and ensure that THOSE don't return new references.

deepCompares are expensive and I really don't want to do them to be called in a critical path such as rendering.  We should instead look to memoize the functions that return these values and only return a new reference if an input has changed.

When you profiled this component, in what situations were we unnecessarily rendering?  Was it related only to dragging items?  Or was it all interactions?

src/lib/Timeline.js
Ilaiwi: Yes, it is for optimization reasons since everytime you call the `head` method a new object for `leftSidebar` and `rightSidebar` will get generated and the `Header` is a pure component which will result in extra renders

src/lib/lines/TodayLine.js
Ilaiwi: Well, this component as well had a lot of extra renders where it got rendered every time out timeline rendered which is not optimal. When I changed it to a pure component this resulted in it being never updated as expected since no state nor props where updated although time changed and it should've moved so it was more of solving a bug. Does this change worth being in a separate pull request?

Ilaiwi:
I agree `deepEqual` and `arrayEqual` are expensive operations but not as expensive as rendering so using them resulted in a much faster and snappier calendar. 
I will look into your suggestion regarding `dimensionItems`, `groupHeights` and `groupTops`. I looked into it the first time and honestly, I don't remember why I didn't do them as mentioned.
Yes, extra renders were caused by all interactions. I really encourage you to install `why-did-you-update` and see for yourself and also try the component after optimization.

mcMickJuice:
Hi @Ilaiwi I really appreciate the effort you put in here however I'm going to close this PR.  I'd like to take a different approach to how this library manages internal state which will change how each component receives props.

I'm definitely going to draw inspiration from these proposals though and make sure to use the why-did-you-update library!

Ilaiwi:
@mcMickJuice you are welcome please let me know if i can help.

Issue #215, opened 2018-01-13
Multiselect feature


mcMickJuice:
Currently have an open PR for this, and a patch to that PR:
* PR - https://github.com/namespace-ee/react-calendar-timeline/pull/131
* Patch - https://github.com/namespace-ee/react-calendar-timeline/pull/198

Will allow for selecting multiple items and dragging them around. Cool!

PR #314, opened 2018-05-23
Refactor multiselect with 0.16.x 

vasdee:
This is a fork of the previous 2 multi-select pull requests but refactored to work with the 0.16.x branch. 

It's not completely ready to merge yet but i'm opening it now for comments and advice. 

- onItemMove now returns a relative unix time to the initial item start time. Would like some advice on whether this is for better or worse as it means it will definitely be a breaking change. 
- canSelect by default enables multi select mode. I was thinking of keeping canSelect as a boolean to allow for turning it off and on and introducing a new prop, multiSelect, (defaulted to false) so that current behaviour of single select is retained by default.

Depending on the above, I will need to update the demos before this can be merged.

- [ ] ~introduce selectMode enum, single, multi, disabled~

- [x] remove canSelect

- [ ] update readme with breaking changes

- [ ] update changelog

- [ ] potential refactor of handleItemMove callback 

- [x] update all demos to work with new changes

- [ ] investigate canSelect override in item props and what changes (if any) need to occur

### DIFF
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -59,7 +59,6 @@ export default class ReactCalendarTimeline extends Component {
     canMove: PropTypes.bool,
     canResize: PropTypes.oneOf([true, false, 'left', 'right', 'both']),
     useResizeHandle: PropTypes.bool,
-    canSelect: PropTypes.bool,
 
     stackItems: PropTypes.bool,
 
@@ -71,8 +70,6 @@ export default class ReactCalendarTimeline extends Component {
     onItemMove: PropTypes.func,
     onItemResize: PropTypes.func,
     onItemClick: PropTypes.func,
-    onItemSelect: PropTypes.func,
-    onItemDeselect: PropTypes.func,
     onCanvasClick: PropTypes.func,
     onItemDoubleClick: PropTypes.func,
     onItemContextMenu: PropTypes.func,
@@ -186,8 +183,6 @@ export default class ReactCalendarTimeline extends Component {
     canMove: true,
     canResize: 'right',
     useResizeHandle: false,
-    canSelect: true,
-
     stackItems: false,
 
     traditionalZoom: false,
@@ -196,8 +191,7 @@ export default class ReactCalendarTimeline extends Component {
     onItemMove: null,
     onItemResize: null,
     onItemClick: null,
-    onItemSelect: null,
-    onItemDeselect: null,
+
     onCanvasClick: null,
     onItemDoubleClick: null,
     onItemContextMenu: null,
@@ -240,7 +234,7 @@ export default class ReactCalendarTimeline extends Component {
     headerLabelFormats: defaultHeaderLabelFormats,
     subHeaderLabelFormats: defaultSubHeaderLabelFormats,
 
-    selected: null
+    selected: []
   }
 
   static childContextTypes = {
@@ -298,8 +292,6 @@ export default class ReactCalendarTimeline extends Component {
       }
     }
 
-    let selectedItems = (this.props.selected) ? this.props.selected.slice() : [];
-
     this.state = {
       width: 1000,
 
@@ -307,7 +299,6 @@ export default class ReactCalendarTimeline extends Component {
       visibleTimeEnd: visibleTimeEnd,
       canvasTimeStart: visibleTimeStart - (visibleTimeEnd - visibleTimeStart),
 
-      selectedItems: selectedItems,
       dragTimeDelta: 0,
       dragGroupDelta: 0,
       infoLabel: null,
@@ -431,12 +422,7 @@ export default class ReactCalendarTimeline extends Component {
       items,
       groups,
       sidebarWidth,
-      selected
     } = nextProps
-
-    if (selected) {
-      this.setState({selectedItems: selected});
-    }
 
     if (visibleTimeStart && visibleTimeEnd) {
       this.updateScrollCanvas(
@@ -615,35 +601,37 @@ export default class ReactCalendarTimeline extends Component {
     )
   }
 
-  selectItem = (item, clickType, e) => {
-    if (
-      this.state.selectedItems.indexOf(item) > -1 ||
-      (this.props.itemTouchSendsClick && clickType === 'touch')
-    ) {
-      if (this.props.onItemClick) {
-        const time = this.timeFromItemEvent(e)
-        this.props.onItemClick(item, e, time)
-      }
-    } else {
+  /* 
+    TODO: the itemTouchSendsClick and clickType logic here 
+    needs more investigation, not sure if required any more...
+    if it is required, it needs to be moved into the new clickItem function
+  */
+  // selectItem = (item, clickType, e) => {
+  //   if (
+  //     this.state.selectedItems.indexOf(item) > -1 ||
+  //     (this.props.itemTouchSendsClick && clickType === 'touch')
+  //   ) {
+  //     if (this.props.onItemClick) {
+  //       const time = this.timeFromItemEvent(e)
+  //       this.props.onItemClick(item, e, time)
+  //     }
+  //   } else {
     
-      if (!this.props.selected) {
-        this.setState({ selectedItems: [item] });
-      }
+  //     if (!this.props.selected) {
+  //       this.setState({ selectedItems: [item] });
+  //     }
 
-      if (this.props.onItemSelect) {
-        const time = this.timeFromItemEvent(e)
-        this.props.onItemSelect(item, e, time)
-      }
-    }
-  }
+  //     if (this.props.onItemSelect) {
+  //       const time = this.timeFromItemEvent(e)
+  //       this.props.onItemSelect(item, e, time)
+  //     }
+  //   }
+  // }
 
-  deselectItem = () => {
-    if(!this.props.selected) {
-      this.setState({selectedItems: []});
-    }
-
-    if(this.props.onItemDeselect) {
-      this.props.onItemDeselect();
+  clickItem = (item, clickType,  e) => {
+    if (this.props.onItemClick) {
+      const time = this.timeFromItemEvent(e)
+      this.props.onItemClick(item, e, time)
     }
   }
 
@@ -721,9 +709,7 @@ export default class ReactCalendarTimeline extends Component {
 
     // if not clicking on an item
     if (!hasSomeParentTheClass(e.target, 'rct-item')) {
-      if (this.state.selectedItems.length > -1) {
-        this.deselectItem();
-      } else if (this.props.onCanvasClick) {
+      if (this.props.onCanvasClick) {
         const [row, time] = this.rowAndTimeFromScrollAreaEvent(e)
         if (row >= 0 && row < this.props.groups.length) {
           const groupId = _get(
@@ -956,20 +942,18 @@ export default class ReactCalendarTimeline extends Component {
         items={this.props.items}
         groups={this.props.groups}
         keys={this.props.keys}
-        selectedItems={this.state.selectedItems}
         dragSnap={this.props.dragSnap}
         minResizeWidth={this.props.minResizeWidth}
         canChangeGroup={this.props.canChangeGroup}
         canMove={this.props.canMove}
         canResize={this.props.canResize}
         useResizeHandle={this.props.useResizeHandle}
-        canSelect={this.props.canSelect}
         moveResizeValidator={this.props.moveResizeValidator}
         topOffset={this.state.topOffset}
-        itemSelect={this.selectItem}
         itemDragStart={this.itemDragStart}
         itemDrag={this.itemDrag}
         itemDrop={this.itemDrop}
+        onItemClick={this.clickItem}
         onItemDoubleClick={this.doubleClickItem}
         onItemContextMenu={this.contextMenuClickItem}
         itemResizing={this.resizingItem}
@@ -1139,7 +1123,7 @@ export default class ReactCalendarTimeline extends Component {
     let dimensionItems = visibleItems.reduce((memo, item) => {
       const itemId = _get(item, keys.itemIdKey);
       const itemTimeStart = _get(item, keys.itemTimeStartKey);
-      const isDragging = isDraggingItem && this.state.selectedItems.indexOf(itemId) > -1;
+      const isDragging = isDraggingItem && this.props.selected.indexOf(itemId) > -1;
       const isResizing = itemId === resizingItem
 
       let dimension = calculateDimensions({
@@ -1159,7 +1143,7 @@ export default class ReactCalendarTimeline extends Component {
       if (dimension) {
         dimension.top = null
         let order = groupOrders[_get(item, keys.itemGroupKey)]
-        if (isDraggingItem && canChangeGroup && this.state.selectedItems.indexOf(itemId) > -1) {
+        if (isDraggingItem && canChangeGroup && this.props.selected.indexOf(itemId) > -1) {
           order = order + dragGroupDelta
           if (order > Object.keys(groupOrders).length - 1) {
             order = Object.keys(groupOrders).length - 1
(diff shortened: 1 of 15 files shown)


mcMickJuice:
This is great. I'll review it this week. 

I'm not very familiar with how move items works.  I also haven't used this behavior in my company's app.  I have a couple of questions:

> onItemMove now returns a relative unix time to the initial item start time. Would like some advice on whether this is for better or worse as it means it will definitely be a breaking change.

What does onItemMove return now as far as time?  From the README:

> Callback when an item is moved. Returns 1) the item's ID, 2) the new start time and 3) the index of the new group in the groups array.

Are you saying you only are returning the time? Or are you returning the item ids of everything that has moved?

vasdee:
> What does onItemMove return now as far as time? From the README:

it's a bit confusing and to be honest I wasn't expecting the existing PR to work this way, but the multi-select onItemMove retains the current signature however, instead of returning the new start time of the item just moved, it now returns the relative time difference from the dragged item's initial start time. If you move the item backwards in time you will get a negative number and conversely moving it forward in time will return a positive number. This time difference can then be applied to all selected objects within your parent component - see the demo for this.

What's also worth noting is that the item param in onItemMove is the actual item that the drag was initiated against, as is the newGroupOrder. It's then up to the parent component to apply the relative time-difference and group order changes to whatever items they are passing in via the selected prop.

> I think it's a good idea to go to a prop that takes an enum, which you noted in the README.

Forgot I left that in the readme actually... I'll make those changes and update the PR

mcMickJuice:
I have another question about how this will function and the UX:

* if i select multiple items and drag them horizontally, I'm assuming the item times will update based on the delta from original X to new X, correct?  So multiselect with changing times will add/subtract from the startTime for each item, correct?  Wouldn't some users assume that multiselect and draggin update all items to the startTime of the items you are dragging?  I.e. the new start time of all items is the new X you drag to?
* Similar to above, when dragging multiple items, would dragging into another group add all the selected items into that target group? Or will it just offset the items into the delta of the groups?  e.g. multiselecting items in different groups and dragging an item from group 4 to group 1 will update all items to change from group X to group X - 3?  (hope that makes sense)

I'm a little hesitant to pursue this change without further discussion with other users of this library. Here's the issue https://github.com/namespace-ee/react-calendar-timeline/issues/215 to discuss this.  

What is your opinion on how this multiselect and drag should behave?

vasdee:
>Wouldn't some users assume that multiselect and draggin update all items to the startTime of the items you are dragging? I.e. the new start time of all items is the new X you drag to?

That was definitely my first assumption when doing this refactor. I think the original author wanted to keep the onItemMove signature somewhat backwards compatible. I'd have to look but it should be possible to get the timeline to perform the updates to the items and return them via onItemMove(). It then might end up looking like:


```
handleOnItemMove(updatedItems: array) {
// Merge the updatedItems back into items props
const items = ...

// Update the selected items if need be 
const selected = ...

this.setState({items, selected})

}
```

> when dragging multiple items, would dragging into another group add all the selected items into that target group?

The demo code explains this better, in the current implementation the user is responsible with reacting to the newGroupOrder and dragTime and feeding that back into the timeline as props. The above proposal would essential remove the following logic ( from the demo) and implement it within the timeline. It would still be up to the user make sure their items/groups state is updated accordingly.

```
handleItemMove = (itemId, dragTime, newGroupOrder) => {
    console.log('item moved');

    const { items, groups, selected } = this.state;

    const group = groups[newGroupOrder];

    // calculate the overall group change for all selected items
    const oldGroupOrder = groups.findIndex( g => g.id == itemId.group);

    // The net offset will be negative for moving up
    const orderOffset = newGroupOrder - oldGroupOrder;

    this.setState({
      items: items.map(item => selected.indexOf(item.id) > -1 ? Object.assign({}, item, {
        start: item.start + dragTime,
        end:  item.end + dragTime,
        group: groups[groups.findIndex( g => g.id ==item.group) + orderOffset].id
      }) : item)
    })
```

andy-dorman:
Initially my vote is for deltas all the way - it’s the only thing that makes sense if you’re changing multiple items at once. I’d have to think about cross group moves but I suspect items should move “between” groups together, maintaining their relative group position. An items group seems fundamentally important to me, also the items it is grouped with, we should try to maintain that integrity where possible.

I’ll try to post something more in depth later today.

vasdee:
> I vote for @vasdee approach

which one? 

1. User has to deal with the dragTime themselves, based on the item that was interacted with (Current Approach)

or

2. Timeline updates the selected items that were moved and returns them as an array. User still has to update their own state to reflect these changes back in the timeline.

mcMickJuice:
I know this is idealistic but I want the user to have as much control over how items are placed, both in groups and time.

I will have time Friday to catch up on this library (there are other pending PRs and issues).  I've been slammed at work.  I'll expand on what I'd like to do as far as how dragging interactions should work, as well as future state of timeline rows as a drag target

vasdee:
>I know this is idealistic but I want the user to have as much control over how items are placed, both in groups and time.

Totally agree. What I've like so much about this library so far is that it does the job of rendering and interacting with the timeline very well.... what happens to items/groups when these events occur should be in the hands of the user.

andy-dorman:
It doesn’t sound idealistic at all. You should aim to keep as much flexibility as possible.

Group squashing should be a flagged option imho. I like that option but I can imagine use cases in either direction. Working with time deltas also seems like the best way to maintain flexibility in combination with an array of item ids.

vasdee:
After playing with the changes for a `selectMode` enum prop, I'm actually leaning towards keeping the canSelect prop to turn the timeline readonly and then removing the internal `selectedItems` state completely and have it controlled via the `selected` prop

The main reason behind this is that if a user wants to use the selected prop, they have to control the selected array from outside the timeline anyway, so why not give them complete control? From what I can see it would simplify some timeline internals, remove duplicate `onItemSelect` and `onItemClick` events and remove assumptions that a user wants their selections to be removed whenever they (accidently) click on the canvas.

Thoughts?

vasdee:
Following on from the above, I've removed the `canSelect` prop and any selected related state from the timeline internals. Now all selected item related logic is driven by whatever the user passes in via the `selected` prop. As a consequence, the onItemSelect/onItemDeselect callbacks are gone as these can now be handled in the user space.

The demo is updated to show how you can replicate the missing internal logic, particularly around selection/deselection of items and how 'deselect' all on a canvas click. 

@mcMickJuice One question I've noted in the source, is around the itemTouchSendsClick in the timeline.js source - can you shed any light on what should happen here?

mcMickJuice:
@vasdee i cant agree more with your last two comments:

* users have complete control over when items are selected/deselected
* onItemSelect and onItemClick are removed.  There are a bunch of headaches around this.

I'm going to play around with your demo of multiselect to see if I can achieve the following:

* move all items from different groups into a single group
* when dragging multiple items with different times, the new startTime is the time of the item you are dragging

This will prove that these changes give ultimate control to the end user in determining the behavior when dragging multiple items.

I am still mulling over these changes and comparing them to other changes I would like to see in this library.  Keep in mind that I took over maintenance of this library at the beginning of the year so I'm not 100% familiar with certain parts of it (specifically, item interaction).

src/lib/Timeline.js
```diff
@@ -606,23 +601,37 @@ export default class ReactCalendarTimeline extends Component {
...
-    } else {
-      this.setState({ selectedItem: item })
-      if (item && this.props.onItemSelect) {
-        const time = this.timeFromItemEvent(e)
-        this.props.onItemSelect(item, e, time)
-      } else if (item === null && this.props.onItemDeselect) {
-        this.props.onItemDeselect(e) // this isnt in the docs. Is this function even used?
-      }
+  /* 
+    TODO: the itemTouchSendsClick and clickType logic here
```
mcMickJuice: I believe this is related to on some touch devices (specifically ios) touch doesn't fire click events unless they have certain attributes:

https://makandracards.com/makandra/34753-how-to-fix-ipad-does-not-trigger-click-event-on-some-elements

Don't worry about this for now.

mcMickJuice:
> * move all items from different groups into a single group
> * when dragging multiple items with different times, the new startTime is the time of the item you are dragging

I can achieve the above with the following code:

```javascript
handleItemMove = (item, dragTime, newGroupOrder) => {
    const { items, groups, selected } = this.state

    // calculate the overall group change for all selected items
    const oldGroupOrder = groups.findIndex(g => g.id == item.group)

    // we need to check to see if target item actually changed groups
    const didChangeGroup = newGroupOrder !== oldGroupOrder
    // dragTime is a delta so calculate the destinationTime from target item
    const destinationTime = item.start + dragTime


    this.setState({
      items: items.map(
        i =>
          selected.indexOf(i.id) > -1
            ? Object.assign({}, i, {
                // start: i.start + dragTime,
                // end:  i.end + dragTime,
                // group: groups[this.calcNewGroupOrder(orderOffset, groups, i.group)].id
                start: destinationTime,
                end: i.end - i.start + destinationTime,
                group: didChangeGroup ? groups[newGroupOrder].id : i.group
              })
            : i
      )
    })
  }
```

This works whens I perform the DROP, however when dragging, the items move around the timeline with their relative offsets:

video - https://imgur.com/a/8RZ9FXE

Let me know if you're seeing the same thing. There are some solutions to this but I'm burnt out after this week and am leaving work soon.

vasdee:
I'm on my phone at the moment and may not get to sit down with this until Monday but just to be clear, is the issue in the video the very last jump where the non-dragged item jumps to the start time of the dragged one?

andy-dorman:
```
I'm going to play around with your demo of multiselect to see if I can achieve the following:

move all items from different groups into a single group
when dragging multiple items with different times, the new startTime is the time of the item you are dragging
```

**Please** don't make this default behaviour.... I can't honestly think of a situation in which these options would be desirable behaviour out of the box...

"group squashing" I could see a potential use for in **some** cases, setting the time of a group of items to the new time of **one** of them I really can't see a use for. If you want to make it an option then go right ahead, that maintains flexibility, but it's not an option I'll be using.

Maybe if you could outline some cases where this behaviour would be useful I could understand where you're going with this? What I would basically like is for items to be able to move back and forth **relative** to each other **and** for groupings to be maintained - so no group squashing, if that was an option I'd be turning it off.

I'm looking at this to help us with an EPG style application, handling episodes of TV shows in groups of "Seasons" for example, or controlling offer windows of a number of different shows, that absolutely should not be easily be switched between channels/groups.

I'm sorry I've not been able to get more involved with the dev, but work is crazy right now. I'm sure if you reduced everything to working with time **deltas** and not absolutes it would all just reveal itself... Passing around absolute values really is just the happiest of happy paths from what I can see and it actually makes the good stuff harder to deal with.

vasdee:
>"group squashing" I could see a potential use for in some cases, setting the time of a group of items to the new time of one of them I really can't see a use for. If you want to make it an option then go right ahead, that maintains flexibility, but it's not an option I'll be using.

I don't think it is intended as default or even an option, @mcMickJuice just wants to see if it is possible given the current state of play. From the looks of his demo gif it looks like he achieved it :)

I very much think the intention is to let the timeline handle the item rendering and interaction, let the user decide where to put those items. These checks are just to make sure that the callbacks are returning enough (and the correct) information so that all these scenarios can be achieved.

One thing that would be good to test is if the moveResizeValidator can be used to prevent group squashing altogether

mcMickJuice:
@andy-dorman i think you misunderstand those comments.  With @vasdee changes, I want to make sure that the user has the flexibility to choose either option.  Which it does. My snippet above was a recipe to move items into the same group/same time when you drag. Your implementation can be different and we can have examples of both approaches.

> Please don't make this default behaviour.... I can't honestly think of a situation in which these options would be desirable behaviour out of the box...
> "group squashing" I could see a potential use for in some cases, setting the time of a group of items to the new time of one of them I really can't see a use for. If you want to make it an option then go right ahead, that maintains flexibility, but it's not an option I'll be using.

I think the goal of these changes is that YOU define the strategy of what happens when items are moved.  I don't think having a default option is the way to go, rather recipes (as mentioned above).

My worry is that this feature is implemented and it makes half of the library's users happy and half upset.

mcMickJuice:
@vasdee the issue in that video is that while dragging, the visual while dragging does not match up with the drop state.  For example, when dragging item A (target item, which is in group 1) and item B (which is in group 3) to group 2, the visual state while dragging is item A is placed in group 2 and item B is placed in group 4 (or 3 if group 4 doesn't exist).  However, when you release (and drop items) both items are placed in group 2, which is the desired state.

I think that moveSizeValidator is absolutely the way to go and it is similar to the [state reducer pattern](https://github.com/paypal/downshift#statereducer) in the downshift library.

andy-dorman:
@mcMickJuice - ok, sounds like you have it covered then..

@vasdee all that behaviour looks good to me, i’ve already started adding keyboard modifiers in my own app so that approach makes sense.

Basically, **I’m** going to need something that mimics audio editing software so all these possibilities are making me happy!

vasdee:
@mcMickJuice How would you like me to move forward with this? 
would `function handleInteraction(...)` replace `moveResizeValidator` in this proposal and `onItemsModified` would replace both `onItemMove` and `onItemResize`?

Internally then, `handleInteraction` function would need to be called wherever `moveResizeValidator` is called now, however now the new `onItemsModified` prop is called when the drag or resize event ends and returns the modifications to all the items ( as specified by  `handleInteraction` )

Let me know if this lines up with your thinking and I can have a crack at it this week.

mcMickJuice:
@vasdee yes, I think that's right,  I do think it should replace `moveResizeValidator` as it would cover that use case and more.  Same with `onItemsModified` replacing `onItemMove` and `onItemResize`.

Maybe we should try to come up with a better name for `handleInteraction`.

Candidates:
* itemsStateFromInteraction
* itemsPositionFromInteraction
* mapItemsStateFromInteraction or mapItemsPositionFromInteraction

What are your thoughts on this?  Do you think these changes will make this library/interaction with items easier to use?  With adequate documentation, do you think this is something that users will appreciate?  Do these changes allow for flexibility and multiple use cases (similar to group squashing vs moving all items to a group, as discussed above)?

vasdee:
@mcMickJuice I haven't spent alot of time looking at this so far this week, but I think it will definitely simplify the internal logic and give more power to the users... at the expense of some out of the box functionality. That can be mitigated somewhat by the docs and recipes you mentioned before.

Naming wise I've gone with `interactionReducer`, at least until I can get something half way working. I'll keep this thread updated with progress, just so it doesn't lose momentum :)

JKCooper2:
Will there be updates for the positioning while dragging as well? Atm I think the collision code only takes into account a single item being moved, so multiple selected items collide with each even if they have been moved. See the gif above when dragging all 4 items after they've been placed in a single group

vasdee:
Possibly, I've only just started getting my head around it so I have no idea what running a state reducer over every drag snap event will do to performance.

If it does make this PR it will be up to the user to write the code in the stateReducer to detect and respond to the collision though

mcMickJuice:
@vasdee I do not agree with user handling collision. I think thats a library concern.  

Just write the reducer to calculate new positions for items.  The library code should handle collisions.  If you need my assistance on this please let me know

JKCooper2:
@mcMickJuice Yes. I'm using a fork of this library at work with multi-select atm, and that's we had issues with that multi drag height positioning, the drag end only doing the item your dragging and not all selected items (not sure if that's been fixed), and the onItemMove and onItemResize firing with different information/a different number of times when you had multiple selected items, so having to be handled differently.

We did get the last one working correctly, I'll try and get some more info on what we had to do.

vasdee:
Quick update from me. I'm no longer focusing on this PR anymore and unless someone wants to pick it up, it can be closed. 

The longer version of the above is that during development of the multi-select branch, the requirement of my project changed such that displaying thousands of activities became the norm. Unfortunately this timeline, for all the great things it does, does not render that many items well. Instead, a couple of us from our project took a lot of learnings and suggestions from this project and created a new timeline built from the ground up with react-virtualised and multi-select out of the box. 

The component has been recently open-sourced on github, however I won't link it here as i'm not sure if that's the proper etiquette. Happy to if @mcMickJuice is ok with it :)

Ilaiwi:
closed for inactivity. If anyone wants to pick this up feel free to open a new PR and I will be happy to help out

PR #325, opened 2018-06-02
Refactor "Horizontal Lines" into Rows component

mcMickJuice:
Simplifies code around determining group you're clicking on.  Also sets the stage for making groups "drop targets" and better management of items in groups.


vasdee:
Question, will this eventually replace the existing plugin system with regards to being able to control the what & how the rows are rendered?

mcMickJuice:
Yes and no.  This change was to simplify the codebase in regards to canvas click handlers and the like.  But this is also a part of a larger effort to change this library's API to be much more composable:
* ability to provide a custom component for HorizontalRow.  So to your question, yes this is where you would be able to customize how the rows are rendered
* ability to attach whatever handlers you want on rows

The plugin system is going to be removed in favor a compositional approach.

Issue #327, opened 2018-06-03
Timeline Markers (Today, Cursor, Custom)


andy-dorman:
I’d like to be able to drop "markers" on to the timeline which would render as a "timeline marker", pretty much like the "Today line" marker works at the moment. I'd be more than happy to handle these outside of the component itself, but if I could build an array of "markers" with corresponding time properties I could do things like set up "ctrl-[NUMBER]" key commands to jump directly to points in the timeline. I suggesting you essentially abstract out the "Today line" marker that gets set when the component initialises so you can create such markers programatically, using them pretty much like bookmarks. So for example, I've already set it up so double clicking on the canvas creates a new item at that point in time, I'd like to be able to create a new "marker" in the same way. Ideally you'd be able to bypass the snapping configuration in many of the use-cases for this stuff, but sometimes you wouldn't, so passing both the rounded values and the actual values would be good.

mcMickJuice:
This is a good idea.  I'd like to implement this with the following goals in mind:

* user maintains state of this feature
* user is in control of rendering of this feature
* api is as declarative as possible and hopefully doesn't introduce new props on the Timeline component.

Here's the end API I have in mind:

```jsx
import Timeline, {TimelineMarker} from 'react-calendar-timeline'

<Timeline>
   // tell the timeline what and how to render TimelineMarker
   {timelineMarkers.map(marker => <TimelineMarker key={marker.id} time={marker.time}>
         // any handlers, or whatever you want in here
       <div></div>
       </TimelineMarker>
    }
</Timeline
```



The `TodayMarker` and `CursorMarker` would be handled in the exact same way:

```jsx
import Timeline, {TodayMarker, CursorMarker} from 'react-calendar-timeline'

// no need to pass in `showCursorLine` anymore
// today line won't appear by default
<Timeline>
  <TodayMarker /> 
  <CursorMarker />

  // can optionally pass in render prop/child as function to change how line is rendered
  <TodayMarker>
  // render whatever you want here
  {(style, time) => <div style={style} />}
  </TodayMarker>
</Timeline>
```

If you want more control over whats rendering, or maybe you wanna attach handlers:

```jsx
<Timeline>
  // can optionally pass in render prop/child as function to change how line is rendered
  <TodayMarker>
  // render whatever you want here...within reason though since timeline does attach certain properties to this element
  {(style, time) => <div style={style} />}
  </TodayMarker>
</Timeline>
```

To your specific use case, I think this approach would work:
[...]

PR #339, opened 2018-06-21
Timeline Markers

mcMickJuice:
**Issue Number**

https://github.com/namespace-ee/react-calendar-timeline/issues/327

**Overview of PR**

Introduces `TodayMarker`, `CursorMarker` and `CustomMarker` feature.  User can pass in these these components as children of the Timeline component.


mcMickJuice:
@andy-dorman would you mind testing this out and seeing if it fits your needs?  You can install this version at 0.18.0-beta.0.

Please, if there are any bugs, file them here and provide as much detail as you can.

Hope you like it! :)

__tests__/index.js
```diff
@@ -34,7 +34,7 @@ const items = [
   }
 ]
 
-describe('Timeline', () => {
+xdescribe('Timeline', () => {
```
Ilaiwi: is there a reason why some test suits are skipped ?

src/lib/Timeline.js
```diff
@@ -1250,103 +1164,100 @@ export default class ReactCalendarTimeline extends Component {
...
+                onScroll={this.onScroll}
+                isInteractingWithItem={isInteractingWithItem}
+                onMouseEnter={this.handleScrollMouseEnter}
+                onMouseLeave={this.handleScrollMouseLeave}
+                onMouseMove={this.handleScrollMouseMove}
+                onContextMenu={this.handleScrollContextMenu}
+              >
+                <div
+                  ref={el => (this.canvasComponent = el)}
+                  className="rct-canvas"
```
Ilaiwi: I see that you removed `.rct-canvas` from scss. Is there a reason why it is there?

src/lib/markers/implementations/CursorMarker.js
```diff
@@ -0,0 +1,107 @@
...
+import PropTypes from 'prop-types'
+import { MarkerCanvasConsumer } from '../MarkerCanvasContext'
+
+const criticalStyles = {
+  position: 'absolute',
+  top: 0,
+  bottom: 0,
+  // FIXME: are these critical styles?
+  width: '2px',
+  backgroundColor: 'pink'
```
Ilaiwi: if these are not critical maybe they should be passed as props and they would have default values as specified here?

src/lib/markers/public/TimelineMarkers.js
```diff
@@ -0,0 +1,20 @@
+// Is this necessary? The initial reason for including this is for organization sake in the
+// user code e.g.
```
Ilaiwi: I really like that you decided to include this. This makes the API more declarative or at least that is what I think

src/lib/scroll/ScrollElement.js
```diff
@@ -114,8 +114,8 @@ class ScrollElement extends Component {
     })
   }
 
-  handleMouseLeave = e => {
-    this.props.onMouseLeave(e)
+  handleMouseLeave = () => {
+    // this.props.onMouseLeave(e)
```
Ilaiwi: is there a reason for not deleting these?

src/lib/markers/implementations/CursorMarker.js
```diff
@@ -0,0 +1,107 @@
...
+  handleCanvasMouseOver = ({ leftOffset, date, isCursorOverCanvas }) => {
+    this.setState({
+      leftOffset,
+      date,
+      isShowingCursor: isCursorOverCanvas
+    })
+  }
+
+  componentDidMount() {
+    this.unsubscribe = this.props.subscribeToCanvasMouseOver(
```
Ilaiwi: maybe call this subscribe? this is confusing to me

src/lib/markers/implementations/TodayMarker.js
```diff
@@ -0,0 +1,66 @@
...
+
+// eslint-disable-next-line
+const defaultRenderer = ({ styles }) => (
+  <div style={styles} data-testid="default-today-line" />
+)
+
+/** Marker that is placed based on current date.  This component updates itself on
+ * a set interval, dictated by the 'interval' prop.
+ */
+class TodayMarker extends React.Component {
```
Ilaiwi: the logic here is very similar to `CustomMarker` and `CursorMarker`. I think maybe extend the `CutomMarkrt` to be the base component and extend its props to take every possible prop that `TodayMarker` and `CursorMarker` could take. This will save a lot of LOC and give the user even more ability to "replicate" the behavior of the other marker components

__tests__/index.js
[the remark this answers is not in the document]
mcMickJuice: yep. It's due to this issue: https://github.com/airbnb/enzyme/issues/1509

This PR uses React 16.3.x which uses the React.createContext api.  This broke these tests to the above issue so I disabled them.

I'm using the create-react-context polyfill so before merging this PR I will revert the react version to 15.x and maintain support for < 16 and reenable these tests.

src/lib/Timeline.js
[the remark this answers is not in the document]
mcMickJuice: Nope, no reason at all. Good catch!

mcMickJuice: Nope, will remove!

src/lib/markers/implementations/CursorMarker.js
mcMickJuice: `styles` are passed to the custom renderer so if the user wants to override width and color they can do so,  I'm still not 100% on this one though.

src/lib/markers/public/TimelineMarkers.js
mcMickJuice: I agree that its more declarative and I could see this component taking props in the future. Just don't tell anyone these markers without using this component :)

src/lib/scroll/ScrollElement.js
mcMickJuice: None. Will delete!

src/lib/markers/implementations/CursorMarker.js
mcMickJuice: are you referring to `this.unsubscribe`?

src/lib/markers/implementations/TodayMarker.js
mcMickJuice: Great point. Will look into that.

mcMickJuice:
Checklist of things to do before merging:

- [ ] Update PR based on @Ilaiwi feedback (thanks!!!)
- [ ] Fix broken tests (revert to local version of React, enzyme 15.x)
- [ ] Revert peer dep update (support 15+)

src/lib/markers/implementations/CursorMarker.js
Ilaiwi: yes

mcMickJuice: `subscribeToCanvasMouseOver` returns a function that, when called removes the subscribed function from the subscription, hence the name, `unsubscribe`.

Ilaiwi: oh okay thanks for clarifying that

PR #308, opened 2018-05-08
Item renderer

Ilaiwi:
Rework ItemRenderer to implement renderProp + propGetters patterns. #289 

- [x] refactor item to use renderProp
- [x] add an example to demo using render props to show the power of the pattern
- [x] edit readme to include the change
- [x] update changelog with breaking changes

renderprop

Issue #244, opened 2018-02-15
Support style for custom coloring


jacwright:
It would be helpful to support the `style` prop. `className` is very useful, but to allow the user to select custom colors for different events you need to be able to set style's `backgroundColor` and `borderColor`.

mcMickJuice:
Does [itemRenderer](https://github.com/namespace-ee/react-calendar-timeline#itemrenderer) prop not support your use case?

jacwright:
No, I don't think so. It only allows the content inside the bubble to be customized, not the bubble itself. At least, that's what it looked like in my testing. Am I wrong?

mcMickJuice:
No, you are correct but you can achieve this through overriding the css classes and itemRenderer prop. left and right padding [is applied](https://github.com/namespace-ee/react-calendar-timeline/blob/master/src/lib/Timeline.scss#L147) to the Item, which is the container in which itemRenderer is rendered.  You'll need to play around with updating/overriding those styles and adding styling to the itemRenderer.

Styles in this library are inflexible and assume too much about use cases.  I created a placeholder story (https://github.com/namespace-ee/react-calendar-timeline/issues/224) to look at a more modern approach to handling styles but I am of the opinion that no styling related to padding or the like should be applied by default.

jacwright:
I will try that out with css and itemRenderer.

buddythumbs:
Hi, did this solve the issue you were having?

I don't know how to change styling in itemRenderer to apply my custom background color to entire item.
This is what I have:
![image](https://user-images.githubusercontent.com/24975408/41792231-649497fc-7658-11e8-9b04-df546153ad56.png)

My item renderer is a styled div:

```
const Item = styled.div`
  width: 100%;
  background-color: ${(props) => props.bg};
  padding-left: 0!important;
  padding-right: 0!important;
  margin-left: 0!important;
  margin-right: 0!important;
`
```
Then the item renderer function:

```
  itemRenderer = ({ item }) => {

     return  <Item bg={item.color}>
        <div >{item.title}</div>
      </Item>
  };
```

Any help here is greatly appreciated.

mcMickJuice:
If you override `.rct-item-content` to remove padding, does that help?  Can you inspect the item element itself and identify _why_ this issue is happening?  I believe it's related to default padding applied.

buddythumbs:
Hey,

[...]

PR #345, opened 2018-06-26
Support Item prop style. Provide prop selected to itemRenderer

morulus:
**Issue Number**

Without issue

**Overview of PR**

Allow to specify custom style for the item element. For example, it can be a different background color for each item (driven by some business logic, etc). Strange that it wasn't implemented before. it is so useful feature.

As well as, provide prop `selected` to the itemRenderer. I have a few cases when I need to know inside the item content does it selected or no. For example, I prefer to render some advanced information inside selected items. Or some way react to this prop in the child component lifecycle - open a modal, or send a request at last.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-calendar-timeline",
-  "version": "0.17.0",
+  "version": "0.17.1",
   "description": "react calendar timeline",
   "main": "lib/index.js",
   "module": "src/index.js",
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres (more or less) to [Semantic Versioning](http://semver.o
 
 ## Unreleased
 
+### Added
+
+* support `Item` prop `style`
+* `selected` is provided to `itemRenderer`
+
 ### 0.17,1
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres (more or less) to [Semantic Versioning](http://semver.o
 
 ## Unreleased
 
+### 0.17,1
+
+### Added
+
+* pass canvasTimeStart/End via timelineContext to the itemRenderer prop
+
 ### 0.17.0
 
 ### Breaking
--- a/README.md
+++ b/README.md
@@ -103,6 +103,9 @@ Expects either a vanilla JS array or an immutableJS array, consisting of objects
   canResize: false,
   canChangeGroup: false,
   className: 'weekend',
+  style: {
+    backgroundColor: 'fuchsia'
+  },
   itemProps: {
     // these optional attributes are passed to the root <div /> of each item as <div {...itemProps} />
     'data-custom-attribute': 'Random content',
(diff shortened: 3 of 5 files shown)


mcMickJuice:
Thanks for opening this PR.

Can please open up an issue describing this PR and provide detail about how the current API doesn't give you what you need?

yftachman:
Can't you achieve this with the itemRenderer?

morulus:
No, itemRenderer gives you access to the inner content of an item, but not to an item itself.

yftachman:
Oh, I see what you are saying.
Yeah this feature is needed, in the mean time you can try overriding the item class, might help you out :) Don't forget to use !important

morulus:
The class is not suitable for me, because I'm in the situation when item background color is formed dynamically by the server. I can not prepare classes for every possible color :)

mcMickJuice:
@morulus can you please open an issue for this PR? No PRs will get merged until there's discussion around its need.

Not saying this won't be added but it overlaps with another issue around refactoring itemRenderer.

morulus:
The `style` issue https://github.com/namespace-ee/react-calendar-timeline/issues/347
The `selected` issue  https://github.com/namespace-ee/react-calendar-timeline/issues/348
Moreover, I found the related issue https://github.com/namespace-ee/react-calendar-timeline/issues/244

morulus:
I want to consult with you. 
Here is a chunk of code, implements style support:

```js
const style = {
      left: `${dimensions.left}px`,
      top: `${dimensions.top}px`,
      width: `${dimensions.width}px`,
      height: `${dimensions.height}px`,
      lineHeight: `${dimensions.height}px`,
      ...this.props.item.style
    }
```

As you can see, It allows overwriting built-in properties. Probably, it is bad, because this can lead to anomalies. And it would be better to do it in this way:

```js
const style = {
      ...this.props.item.style
      left: `${dimensions.left}px`,
      top: `${dimensions.top}px`,
      width: `${dimensions.width}px`,
      height: `${dimensions.height}px`,
      lineHeight: `${dimensions.height}px`,
    }
```

Thus, the built-in properties will be kept. How do you think?

Ilaiwi:
All the usecases for this will be supported when #308 gets merged but  until then this will be useful.

mcMickJuice:
Agree with @Ilaiwi .  I might merge this PR but these use cases will be achievable when #308 is merged.  I might prioritize that PR though as it's the direction we want to go anyways.

morulus:
Ok. I've updated pr according to https://github.com/namespace-ee/react-calendar-timeline/pull/345#issuecomment-400348465. Changed the order of style in the spread. https://github.com/namespace-ee/react-calendar-timeline/pull/345/commits/caf2cdaa4eee2cb2b7ab7710bcefab463a501ee7

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-calendar-timeline",
-  "version": "0.17.0",
+  "version": "0.17.1",
```
mcMickJuice: For future PRs, don't worry about updating version here. I'll update that as a part of a release

mcMickJuice:
re: https://github.com/namespace-ee/react-calendar-timeline/pull/345#issuecomment-400348465

Yes absolutely, these styles should be overridden by any styles that the timeline is applying (i.e. left, top etc)

package.json
morulus: Oh, I did it, really? Sorry, just a habit)

Issue #352, opened 2018-06-28
PeerDependency version of interactjs doesn't work with browserify


yftachman:
**Describe the bug**
Hey, since moving interactjs as peer dependency I can't run my project unless I delete the dist folder of interact js. 
Obviously this is their bug however this post: https://github.com/taye/interact.js/issues/593
has a solution that might help as a workaround in the mean time.
   
**To Reproduce**
1. npm install react-calendar-timeline
2. WARN have peer dependency interactjs
3.  npm install interactjs
4. project can't run because  
"Error: Cannot find module './src/utils/window' from '/Users/.../git/app/node_modules/interactjs/dist'"

Not sure what to do as making all of my team mates  delete the dist folder of a some random library sounds pretty ridiculous.

mcMickJuice:
@yftachman thanks for opening this. Would you mind submitting a small PR to fix this? interactjs is in Item.js file

mcMickJuice:
I'm going to close this as this is an issue with interactjs library, not this one.

PR #353, opened 2018-06-28
fixing the import of interactjs (#352)

yftachman:
**Issue Number**
#352 

**Overview of PR**
I wasn't able to really test this except I know from messing around in my own separate project, that making this changes fixes the problem. 
I'm not sure why, I think the issue is my project uses browserify and this one is using webpack (maybe webpack knows how to handle this). But yarn start works fine.

### DIFF
--- a/src/lib/items/Item.js
+++ b/src/lib/items/Item.js
@@ -1,6 +1,6 @@
 import React, { Component } from 'react'
 import PropTypes from 'prop-types'
-import interact from 'interactjs'
+import interact from 'interactjs/src/index';
 import moment from 'moment'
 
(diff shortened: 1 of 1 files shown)


mcMickJuice:
I'm not going to merge this as this is not this library's concern. I might remove interactjs anyways.

yftachman:
I guess this library is unusable for me and anyone else who uses browserify until then.

yftachman:
Maybe it's just this version of interactjs, do you know if this is the minimum version of interactjs this library depends on having? Is there a lower version I can install that will still support this library?

If so, we can lower the minimum version in our peer dependency and anyone with this problem can just put in their peer dependency the lower version instead of latest

Issue #377, opened 2018-07-24
groupRenderer doesnt work for right sidebar


Ilaiwi:
**Describe the bug**
Timeline's right sidebar doesn't content doesn't render the `groupRenderer` passed to the timeline

**To Reproduce**
1. provide `rightSidebarWidth` prop to timeline
2. provide `groupRenderer` prop to timeline

**Expected behavior**
right sidebar content should be rendered through `groupRenderer` if provided

**Library Version**
0.18

**Desktop (please complete the following information):**

* OS: macOS
* Browser Chrome

**Additional context**
fixing this is easy just pass `groupRenderer` to the right sidebar 
[code](https://github.com/namespace-ee/react-calendar-timeline/blob/master/src/lib/Timeline.js#L920)

mcMickJuice:
Fixed in 0.18.1

PR #381, opened 2018-07-24
fix grouprenderer not affecting right sidebar

Ilaiwi:
**Issue Number**

#377 

**Overview of PR**

- [ ] add test


mcMickJuice:
@Ilaiwi have you tested this locally? I don't think this needs a unit test. It seems like a basic oversight.

Ilaiwi:
yes I have it running on some alpha

PR #385, opened 2018-07-26
Bringing sensible `scroll-behavior` value to prevent aberrant behavio…

fguitton:
This is self explanatory. When `scroll-behavior` is set to a different then 'initial' value, the timeline behaves in erratic fashion. To reproduce set the default scolling behaviour of all element to `smooth` and you will see the timeline missbehaving badly.

Setting this value explicitly to `unset` (rather than `inital`) is more sensible and creates a safe default.

### DIFF
--- a/src/lib/Timeline.scss
+++ b/src/lib/Timeline.scss
@@ -23,6 +23,7 @@ $weekend: rgba(250, 246, 225, 0.5);
 .react-calendar-timeline {
   * {
     box-sizing: border-box;
+    scroll-behavior: unset;
   }
 
   .rct-outer {


mcMickJuice:
Thanks for opening this issue.

Can you create an issue for this with example of with and without?  Generally it's good practice to open an issue prior to a PR to discuss.  I know this is very simple but now the onus is on me to do A B testing of this.  I'm sorry but I don't have time for that.

Just a minimal example will suffice and I'll review and merge.

mcMickJuice:
No response after comment from a week ago.  This is pending an issue to discuss why this is needed.

fguitton:
@mcMickJuice I have been quite busy by other things this past week, I will come back to this when I get a bit of time. There are multiple use cases we are having that are requiring some rather uncomfortable hacking of the component. I will also try to bring those forward.

Issue #332, opened 2018-06-10
Can't scroll vertically if inside a fixed div


yftachman:
If any of the containing elements have position fixed I'm no longer able to scroll vertically.
Any ideas how to get around it?

Example:
https://codesandbox.io/s/rlw48868jo

(Or just add position:fixed to any of the containing elements)

Happens on latest (0.16.3)

yftachman:
Found the culprit.
In the ScrollElement: 
`  handleWheel = e => {
    const { traditionalZoom } = this.props

    e.preventDefault()

    // zoom in the time dimension
    if (e.ctrlKey || e.metaKey || e.altKey) {
      const parentPosition = getParentPosition(e.currentTarget)
      const xPosition = e.clientX - parentPosition.x

      const speed = e.ctrlKey ? 10 : e.metaKey ? 3 : 1

      // convert vertical zoom to horiziontal
      this.props.onWheelZoom(speed, xPosition, e.deltaY)
    } else if (e.shiftKey) {
      // shift+scroll event from a touchpad has deltaY property populated; shift+scroll event from a mouse has deltaX
      this.scrollComponent.scrollLeft += e.deltaY || e.deltaX

      // no modifier pressed? we prevented the default event, so scroll or zoom as needed
    } else {
      if (e.deltaX !== 0) {
        if (!traditionalZoom) {
          this.scrollComponent.scrollLeft += e.deltaX
        }
      }
      if (e.deltaY !== 0) {
        **window.scrollTo**(window.pageXOffset, window.pageYOffset + e.deltaY)
        if (traditionalZoom) {
          const parentPosition = getParentPosition(e.currentTarget)
          const xPosition = e.clientX - parentPosition.x

          this.props.onWheelZoom(10, xPosition, e.deltaY)
        }
      }
    }
  }`

I think the way to solve this is to get a containing element from the user (If none provided use the window). Any thoughts?

acemac:
When using a fixed div, drag and drop is broken if the div is scrolled.

mcMickJuice:
This is likely due to the fact that drag and drop event is looking at PageY which might be screwed up if you use a fixed div.  Just a guess.

yftachman:
I'm having difficulty making this work as expected, anyone with knowledge on how the dragging works can shed some light on the subject?

PR #390, opened 2018-07-29
332 vertical scroll in fixed div

yftachman:
**Issue Number**
https://github.com/namespace-ee/react-calendar-timeline/issues/332

**Overview of PR**
Added a property called verticalScrollContainer, if provided referring to that when wanting to scroll or calculate the scrollTop. If not provided falling back to the window element (like we have today)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres (more or less) to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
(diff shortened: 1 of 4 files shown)


tobiasbueschel:
hey @yftachman & @Ilaiwi, is this PR still relevant or has it been fixed in the meantime? I'd love to get this merged in :-)

Let's also add some tests so we prevent against this bug going forward.

Ilaiwi:
@tobiasbueschel  I think the issue is still relevant but not sure about the solution.
could you resolve the conflict and try it please?

Issue #356, opened 2018-06-29
Header Dates are incorrect if groups is initially empty


morulus:
If as an initial value for prop `groups` I pass an empty array, then the displayed date does not correspond to the values of `defaultTimeStart` and `defaultTimeEnd`.

This causes problems with lazy loading of groups and when empty groups came from the server

Here is the example. I'm trying to display the year 2018, but 2017 is displayed.
https://codesandbox.io/s/rwz71pp20n

mcMickJuice:
This is a nasty bug to fix so until I can get to it, the workaround is, instead of an empty array, pass in a dummy group, maybe with a title of "fetching groups" or something.  Then your default start/end will be respected and everything will be fine when your actual groups are loaded.

As an FYI, the bug is related to how the Scroll header is affected when you scroll the timeline body.  When you have no groups, you have no timeline body so there is no initial scrollLeft offset reported to the header.

morulus:
Ok, I got it.

Ilaiwi:
released in 0.25.0

Issue #357, opened 2018-07-02
Timeline scroll causes Timeline component to re-render on every scroll


mcMickJuice:
**Describe the bug**
When you scroll the timeline, setState is called in the `onScroll` handler.  This causes Timeline component to re-render on every scroll event which can be a performance issue.

**To Reproduce**

Run this repo locally and put a console.log in the render method of Timeline component

**Expected behavior**
Timeline should only re-render when necessary (props change, state changes,  etc

**Screenshots**

**Library Version**
Latest - 0.17.2

mcMickJuice:
I think we'll need to refactor how this library handles scroll.

Current state:
* `ScrollElement` calls `this.props.onScroll` any time the user scrolls that component (swipe, arrow key, drag).
* `Timeline` component's onScroll is invoked with leftOffset.  It calculates the new visibleTimeStart/End and calls `this.props.onTimeChange`. `this.setState` is then called to set offsetLeft.
* `onTimeChange` [allows users to restrict date ranges](https://github.com/namespace-ee/react-calendar-timeline#ontimechangevisibletimestart-visibletimeend-updatescrollcanvas) by calling `updateScrollCanvas`
* `updateScrollCanvas` is called to to determine if the canvas should be reset (i.e. you've scrolled too far to the left or right and we need to "reset" the canvas).  `setState` is called with the new visibleTime values along with canvas (if canvas is reset), items, etc etc.

Multiple set state calls are made for each scroll. We should identify which values are crucial to rendering this timeline and which are not.  IMO `visibleTimeStart/End` and `currentScrollLeft` do not determine what is rendered (`canvasTimeStart/End` do) and therefore should not be involved in `setState`.https://blog.kentcdodds.com/the-state-reducer-pattern-%EF%B8%8F-b40316cfac57
In addition, `this.onTimeChange` is overloaded and should NOT be passed the`updateScrollCanvas` function.  If the purpose of passing updateScrollCanvas to a user is to restrict the current time range, we can introduce a min/max set of props, or something similar to a [stateReducer](https://blog.kentcdodds.com/the-state-reducer-pattern-%EF%B8%8F-b40316cfac57).

acemac:
Is `updateScrollCanvas` neccessary as a callback from `props.onTimeChange`? Can the controlling component achieve the same functionality by passing `visibleTimeStart `and `visibleTimeEnd` based on the result?

mcMickJuice:
[...]

PR #392, opened 2018-07-30
Remove unncessary setstate

acemac:
**Issue Number**

#357 

**Overview of PR**

Timeline should not cause as many setState changes while scrolling.

- Removed the header scroll from state in favour of using references
- Stopped guarantee of setState being called on scroll (will still call if canvas needs to be updated)
- Refactored Header to not have sidebar headers in them so sidebarContent is not re-rendered on scroll or canvas update (our sidebarContent is expensive to render)


mcMickJuice:
@acemac can you please change the base branch to develop?

acemac:
Rebased to develop

Refactored RCT to use React 16.3 new life cycle methods and updated the peer dependency. Didn't gain a great deal from this so maybe unnecessary?

I'm unsure on the style I'm setting in
TimelineElementsHeader line 240
`style={{ height: twoHeaders ? headerLabelHeight : headerLabelHeight + headerLabelGroupHeight, width: canvasWidth }}`

acemac:
Did a bit of clean up and merged develop back in

This fixes the bug with header scrolling when the timeline does not.

Ilaiwi:
Hi @acemac Thank you for your effort. I merged tried this in my very heavy timeline and it gave me a good performance boost. 
I noticed that you changed how the sidebar is rendered to render the sidebar header with it and give the sidebar the height of the full timeline. Could you please elaborate on why this was the way to go? Maybe it is not me but I think the current abroach of having divided the timeline into header and calendar and in the future maybe the scheduler could render multiple timeline headers above, under or anywhere in the view.

Please let me know what do think. And thanks again for the effort

acemac:
Originally I intended to move the header in to the scroll canvas. So broke out the timeline's header to allow for it. I had trouble doing so because `position: sticky` is a nightmare.

I imagine the whole component as three vertical columns, with the middle one doing the scrolling. I also like it because:
- Keeps the sidebar headings from being the wrong size and the timeline heading being off
- Keeps the re-rendering of the timeline heading away from re-rendering the sidebar heading (which takes ages for us, the heading is full of filters, controls, draggable areas etc).

In the future I may end up allowing for replacement of the entire sidebar and heading to be a single component that is provided the heights and tops of each group. And this change is closer to that.

I'm also not particularly attached to how I changed it to be and I can have a go at changing it back.  Has it broken something for you?

Ilaiwi:
Thanks for your fast response. Sorry for the very poor English on my comment, but I was half asleep.

I get what you are doing here, and actually, this was the case in `0.15` but @mcMickJuice opt out of this in his refactoring. I think, maybe we think of the component in a different way. 

So recently, I started working on adding custom headers for the component and I base my implementation on the fact that the header is a separate piece than the timeline itself. This, in my opinion, will give the user the ability to render as many headers as they want, anywhere in their views. For example, A user might render some custom header below the timeline with custom scrollable header and with the sidebar headers. 

Moreover, This will make `header sticky` the user responsibility which I think should be the case. I actually deleted `position: sticky` in the new `itemRenderer` to be a style provided from the component and just have it now in the default item renderer and if the user needs it, then they should handle adding it since they can do it easily

I think this PR should maybe focus on performance and any changes on how the component is divided could be handled in a different issue? 

Please let me know what do you think and if this has a direct relation with performance.

acemac:
This fixes #356

mcMickJuice:
@acemac thanks for your work on this! Would you mind updating the broken tests and letting me know when this is good to go so that I can review?

acemac:
@mcMickJuice fixed the tests. Good luck reviewing the spaghetti code :)

src/lib/Timeline.js
```diff
@@ -343,6 +342,76 @@ export default class ReactCalendarTimeline extends Component {
...
+    const {
+      visibleTimeStart,
+      visibleTimeEnd,
+      items,
+      groups
+    } = nextProps
+
+    // This is a gross hack pushing items and groups in to state only to allow
+    // For the forceUpdate check
+    let derivedState = {items, groups}
```
Ilaiwi: well, I understand why you did this but I am not quite sure this is the way to go. I don't have enough knowledge in `getDerivedStateFromProps`, but I think this lifecycle method is enforced by react not to do hacks like this.

I am not very sure what would be the alternative since I am not very well educated on how to use this life methods more than the couple of posts I read about them. I think we should revert back to `componentWillRecieveProps` and migrate to this later on. 

I am interested in knowing what do you think? and I am interested in @mcMickJuice  opinion as well.

Ilaiwi:
@acemac Thank you for the great work! I've been using this for the past month and I haven't got any problems using it and the performance gain is great!

Sorry for the late code review but I was really busy at work. I added some comments and a discussion about `getDerrivedProps`. Let's discuss it and let's merge this into master soon!

src/lib/Timeline.js
Ilaiwi: sorry approved this by mistake

src/lib/Timeline.scss
```diff
@@ -64,6 +64,7 @@ $weekend: rgba(250, 246, 225, 0.5);
   .rct-header-container {
     z-index: 90;
     display: flex;
+    overflow: hidden;
```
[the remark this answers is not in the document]
acemac: Probably no reason and it could be removed without hurting anything.

acemac:
The getDerivedStateFromProps is a frustrating lifecycle method. The way I've done it, is the way they intend it to be used. But I do not like it.

[See here](https://github.com/facebook/react/issues/12188#issuecomment-364273107)

src/lib/Timeline.js
acemac: The getDerivedStateFromProps is a frustrating lifecycle method. The way I've done this, is the way they intend it to be used.

From React 16.3, `componentWillReceiveProps` is renamed `UNSAFE_componentWillReceiveProps` and this is apparently the way to go.

[See here](https://github.com/facebook/react/issues/12188#issuecomment-364273107)
[and here](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops)

Issue #384, opened 2018-07-24
Stackable items


LitvinaviciusM:
Hey guys, 

Not sure if it is a feature request yet.
Is there any way to stack items only for a specific group ? 

Thanks for a great lib! Any advice is welcome!

mcMickJuice:
I don't believe this is currently possible, although someone else might want to chime in if they were able to achieve this.

Ilaiwi:
This would be very helpful for me as well!

Ilaiwi:
released in 0.23.0

Issue #367, opened 2018-07-17
Custom styles for horizontal lines (groups)


dkarnutsch:
**Is your feature request related to a problem? Please describe.**
Especially when using the tree groups example it is frustrating that I can't style the headers of the groups different than the groups itself. Also I would like to overwrite the calculated height with a custom height.

**Describe the solution you'd like**
For the height: If the height is specified in the group, the lib should use this as the height and not the calculated height.

For the custom style: Every time the horizontal line is rendered, I would like to have a callback where I can supply custom styles. 

**Describe alternatives you've considered**
I've seen that there was/is a plugin approach which makes it possible to make something of this happen, but it seems that there is no documentation and the link to the example is dead. Alternatively we could support categories natively. 

**Additional context**
I will submit an PR.

mcMickJuice:
Released in 0.19.0

PR #397, opened 2018-08-02
Can stack items by group

acemac:
**Issue Number**

#384 

**Overview of PR**

- Added optional prop to group that overrides stackItems
- Refactored stacking of items to listen to new property


mcMickJuice:
@acemac thank you for doing this work. Can you explain what you're doing here?  There are few code comments and it's very hard to follow these changes.

To be honest, I'm not sure if I want to merge this change:
* yes, someone opened an issue that this PR fixes, but how many other users would utilize this feature?  Are we expanding an already bloated API
* are there better ways to handle stacking and item ordering than the current implementation?  If we're going to refactor stacking, why not do a whole refactor and improve the codebase?
* based on a glance over the code, it looks like you're specifying stacking behavior using a property in the group object? Is that right? If so, that's further conflating behavior with data, something this library IMO does too much of.  Why not have a function that's passed in to the timeline that is called for each group and is separate from the group/item data

This is a discussion that should occur on the issue, and not after a PR has been raised. Again, I appreciate the work you're doing on this but to save you some time, we should really talk through it before you open a PR.

acemac:
Well said @mcMickJuice, this is something we will use internally so I was happy to contribute. It however not a very useful function and provides little. Our backend will enforce no overlaps in all but one case. calculating impossible stacks may degrade our front end performance.

Previously in calendar.js there were two functions called stack and nostack. A great deal of the code within them was copy paste. I have merged them in to a single function that checks on the stackItems prop of the timeline, and the new optional stackItems prop on group used to override. Whatever the existing logic did, it still does... you're right about the comments.

There was also some funny array looping going on that I've hopefully made more sensible, previously it was assumed group.id would be sequential and the array is constructed as such. group ids of 1,2 and 9999 would have created an array of length 10000 which is looped over.

The collision detection for items was also unnecessarily checking items that it could not possibly collide with as they were yet to be placed on the timeline.

If you think this PR has merit I can attempt to comment more thoroughly, otherwise happy for you to leave it out.

acemac:
@mcMickJuice I think this PR is worth adding. 

I have merged changes by @dkarnutsch to provide the custom group height. It was much easier to achieve what the intent of this change was.

`calendar.js` `getGroupOrders` is a great place to calculate derived properties for each group, such as custom group `height`, and `stackItems` etc. It is only called once and called right before the item positions are placed on the timeline.

Ilaiwi:
@acemac could you please add the issue @dkarnutsch opened for custom heights?

acemac:
It was done for #367 with PR #368

Ilaiwi:
@acemac could you please fix the conflict with develop?

acemac:
@Ilaiwi Done!

Taken from [my develop branch](https://github.com/acemac/react-calendar-timeline/tree/develop)

Ilaiwi:
@acemac Thanks! I will get to all your PRs ASAP, but I am slammed with work these days.

Ilaiwi:
@acemac So i've been testing this for personal usage and it is not working quite well for me when I have the timeline `stackItems=false` and a group with `stackItems=true`.
I am looking into this right now but maybe you know where is the problem

acemac:
@Ilaiwi thanks!

I can't commit until Monday (comp is dead) but I will tag you where the error is.

I've also noticed I've stuffed this and merged PR #408 in to this PR too.

src/lib/utility/calendar.js
```diff
@@ -454,14 +468,13 @@ state
...
-  const stackingMethod = stackItems ? stack : nostack
-
   // Get a new array of groupOrders holding the stacked items
-  const { height, groupHeights, groupTops } = stackingMethod(
+  const { height, groupHeights, groupTops } = stackAll(
     dimensionItems,
     groupOrders,
     lineHeight,
-    groups
+    groups,
```
acemac: @Ilaiwi This is the error, The groups argument is not needed for stackAll as it now comes in with groupOrders

src/lib/utility/calendar.js
```diff
@@ -247,110 +266,105 @@ export function collision(a, b, lineHeight, collisionPadding = EPSILON) {
...
-
-    // calculate new, non-overlapping positions
+  for (var index in groupedItems) {
+    const groupItems = groupedItems[index]
+    const {items, group} = groupItems
     groupTops.push(totalHeight)
-
+  
+    // Is group being stacked?
+    const isGroupStacked = group.stackItems !== undefined ? group.stackItems : stackItems
```
acemac: @Ilaiwi This is where the logic for the group stacking is decided. The groups array is truthy.

Ilaiwi:
@acemac thanks for the fast response this saved me so much time! I have another bug though where I cant drag items to the `stackItems` group. I will debug this and open a PR with the fixes in your fork so we can both save time.

acemac:
@Ilaiwi , I was able to drag items in groups where  globally `stackItems={false}`, with groups `stackItems={true}`

If you were testing using demo-main/index.js, there is a moveResizeValidator introduced with #408 that ended up in here because I did some terrible merges to these branches. The validator does a really budget job of preventing certain moves on line 133

Ilaiwi:
@acemac Thank you everything is working fine now. Could you please remove API changes from this PR?

Ilaiwi:
replaced by another PR

Issue #408, opened 2018-08-19
Change onItem* API to return Item


acemac:
**Is your feature request related to a problem? Please describe.**
When an item is dragged, editted, moved etc. The `itemId` is passed to the callback.
For simple modifications as in the examples this works well with the `setState` and `map`. But for more complex cases this isn't necessarily ideal.

Every edit, the first thing we have to do it search the items array for the matching id. We take this, build a clone of the object and push it in to changed state array to allow for undo/redo.

**Describe the solution you'd like**

- Add `item` as a fourth argument to the `onItem*` callbacks. The item is available to the origin of the call
- Change documentation of `onCanvas*` methods to have `groupId` instead of `group`
- Add `groupId` to `onItemMove`, the `newGroupOrder` index is the only place the index is used.

**Describe alternatives you've considered**
Search the items array every time

**Additional context**
I am happy to do this work and submit a PR.

Ilaiwi:
Thank you. This will make the API much nicer. For `onItemMove` let's have the group as well passed instead of the group index to make everything consistent

mcMickJuice:
This would cover https://github.com/namespace-ee/react-calendar-timeline/issues/237

developer-ocansey:
Closing as part of issue triage. This issue was created before 2023 and has had no recent activity. If you're still experiencing this on v0.30+, please open a new issue with a minimal reproduction.

Issue #401, opened 2018-08-10
Scroll left/right causes item move/edit to be at incorrect time


acemac:
**Describe the bug**
When moving or resizing a booking while scrolling with shift+mousewheel, the position of the action is wrong

**To Reproduce**
1. Go to https://codesandbox.io/s/6vo2jkov23
2. Click on any booking and move it left/right with the mouse
3. Hold shift+mousewheel
4. Booking position will no longer be under mouse

**Expected behavior**
Booking should move with the scroll, bookings can then be moved past the screen/canvas by scrolling

**Library Version**
19.0

**Desktop (please complete the following information):**

* OS: Win10
* Browser Chrome
* Version 67

Ilaiwi:
released in 0.22.0

Issue #338, opened 2018-06-21
Performance Improvements V2


mcMickJuice:
Let's discuss additional performance improvements that we can make to this library.  Here are a couple of my goals and also some that others have mentioned:

* lazy rendering - only render items that are in or near viewport.  This can be achieved by having an observer on each group row. If that group row is in or near viewport, then we render items in that group. If not, then we don't.  This has some drawbacks as you don't have ctrl + F functionality but we can probably get around that
* cut down on unnecessary renders - don't rerender on canvas mouse over when `showCursor` is true
* Utilize PureComponent

jabidof:
Have a look at https://github.com/NoamELB/shouldComponentUpdate-Children/blob/master/README.md
Not the canonical approach I would like to see implemented (from an academically viewpoint) but I could get a performance boost of a factor of 6-8! 
For just a one-liner I adopted it 👏

Ilaiwi:
an improvement would be to memorize the results from the `dimensionItems`, `groupHeights`, `groupTops` this will of course cut down extra renders as well...

jabidof:
Any chance react waypoint would improve performances?
http://brigade.github.io/react-waypoint/

mcMickJuice:
For sure @jabidof .  I was looking at that or using a polyfill for [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API).  Either way, I think my strategy for this would be:

* use two waypoints as a top and bottom "sentinel". These sentinels will be tied to a GroupRow so we will know what groups are within visible range (+ / - some padding)
* For those groups that are within visible range, only render Items tied to those groups (vertical culling)
* as you scroll, these waypoints will be hit and reset, which will update the range of visible groups

This is theoretical but that should cut down on rendering of items that are out of view.  The same approach can be applied horizontally.

mcMickJuice:
This is a big one too: https://github.com/namespace-ee/react-calendar-timeline/issues/357

acemac:
Would https://github.com/bvaughn/react-virtualized be possible for this?

Or any hints? I'm going to end up with thousands of groups and will need to cut down on re-render times for them.

Ilaiwi:
Yes, it would. actually, I might experiment with it in the following two days

Shvarts:
@Ilaiwi Did you have any success on experimenting with https://github.com/bvaughn/react-virtualized?

Ilaiwi:
[...]

PR #412, opened 2018-08-29
Api update

acemac:
**Issue Number**

#408 

**Overview of PR**

Added argument to the API calls to provide more complete information.
All arguments added to end so should not be a breaking change

### DIFF
--- a/src/lib/items/Item.js
+++ b/src/lib/items/Item.js
@@ -251,7 +251,10 @@ export default class Item extends Component {
             dragTime = this.props.moveResizeValidator(
               'move',
               this.props.item,
-              dragTime
+              dragTime,
+              undefined,
+              this.props.order + dragGroupDelta,
+              this.props.order
             )
           }
 
@@ -259,7 +262,8 @@ export default class Item extends Component {
             this.props.onDrag(
               this.itemId,
               dragTime,
-              this.props.order + dragGroupDelta
+              this.props.order + dragGroupDelta,
+              this.props.item
             )
           }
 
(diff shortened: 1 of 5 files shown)


mcMickJuice:
@acemac would you like to be added as a maintainer of this repo?

acemac:
@mcMickJuice I'm happy to, but I may not stick around to help maintain other than keeping the typescript definitions up to date at [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped) 

We're using this internally for a project that ends next year. In future sprints I will include fixing #401 and will do #338 if we have performance issues. #332 / #390  may get finished by us if we have to.

mcMickJuice:
Fair enough dude! Your contributions are greatly appreciated either way.

acemac:
I've changed how moveResizeValidator works to allow for modifying of the group. Previously it return a `number`, it now returns:
`number | { time: number, newGroupIndex: number }`

By returning the object, the time variable is treated how the existing return was, and the newGroupIndex changes the group index of the return is used to set the group index.

I'm not in love with this as a solution because the index is a bad way to go about it. But I need it currently.

Ilaiwi:
@acemac this will be released in 0.22.0 during the next two weeks thank you for your work

Ilaiwi:
I am gonna push this to the next release because I want to get #401 without huge breaking changes

tobiasbueschel:
Hey ya, just wondering if there is anything I can help with to make progress on this PR? :-) Would be great to add some tests as well.

Ilaiwi:
@tobiasbueschel 
yes we need some help on this.
- resolve conflict
- add tests
I would appreciate the help

src/lib/items/Item.js
```diff
@@ -248,18 +249,31 @@ export default class Item extends Component {
...
+              this.props.order + dragGroupDelta,
+              this.props.order
             )
+            
+            if (validResult.time) {
+              dragTime = validResult.time
+            } else {
+              dragTime = validResult
+            }
+            if (validResult.newGroupIndex) {
```
megasus: If newGroupIndex is set to 0, this statement will return false.
Should be more like:
`if (validResult.newGroupIndex !== undefined)`

src/lib/items/Item.js
```diff
@@ -273,19 +287,34 @@ export default class Item extends Component {
...
+                this.props.order + dragGroupDelta,
+                this.props.order
               )
+
+              if (validResult.time) {
+                dragTime = validResult.time
+              } else {
+                dragTime = validResult
+              }
+              if (validResult.newGroupIndex) {
```
megasus: see previous comment

Issue #421, opened 2018-09-14
When CustomMarker date prop changes marker doesn't move on the timeline


kevinmanncito:
**Describe the bug**
Changing the `date` prop passed into a `CustomMarker` component doesn't update on the timeline.

**To Reproduce**
Code sandbox: https://codesandbox.io/s/q9x8zp0qp9

Just enter in numbers in the text field above the time and you can see that the state is updating but the marker on the timeline doesn't move.

~Run the `marker-change-bug` branch on my [fork](https://github.com/kevinmanncito/react-calendar-timeline/tree/marker-change-bug).~

~Go to the main demo and add numbers in the text field I added and you can see that the state of the marker date is changing but the timeline isn't updating.~

**Expected behavior**
The timeline marker should update if the date being passed into the timeline marker changes.

**Screenshots**
![marker_bug](https://user-images.githubusercontent.com/2521298/45572422-753cff80-b826-11e8-8daa-fe36fe3a2d01.gif)

**Library Version**
0.19.0

**Desktop (please complete the following information):**

* OS: macOS 10.13.6
* Browser: Chrome 69.0.3497.92

**Smartphone (please complete the following information):**
N/A

**Additional context**
Hopefully I'm not missing something obvious but I've been looking at this for a while and it feels like it should be updating. Looks like it could be a bug with the markers provider/consumer logic. Not sure how to fix it tho. Any ideas? Thanks!

Ilaiwi:
Hi, @kevinmanncito could you please replicate this in https://codesandbox.io/s/zr5r289rm4
It will be easier for all of us helping you here

kevinmanncito:
Updated issue with a code sandbox demo: https://codesandbox.io/s/q9x8zp0qp9

Ilaiwi:
@kevinmanncito yes this is a bug. The problem is that the `CustomMarker` inside `src/lib/markers/implementation/CustomMarkers` is not getting the new date. So there is a quick hack fix that will work just fine implemented [here](https://codesandbox.io/s/6x22l3npkz)
I would appreciate if you went and fixed submitted a PR and fixed this issue 🥇

kevinmanncito:
You got it: https://github.com/namespace-ee/react-calendar-timeline/pull/422 let me know if you can see a better way of fixing this

Ilaiwi:
released in 0.20.0

PR #422, opened 2018-09-15
Updating custom marker's date if date changes

kevinmanncito:
**Issue Number**
[421](https://github.com/namespace-ee/react-calendar-timeline/issues/421)

**Overview of PR**
When the date prop on the public `CustomMarker` changes the `markers` array in the `TimelineMarkersProvider` state also needs to change. The fix here is just calling the unsubscribe from the public `CustomMarker` when the date changes and then calling the `subscribeMarker` func again.  The change in date is detected by using the `componentDidUpdate` lifecycle method.


Ilaiwi:
Thank you for this and thanks for your first contribution. I think this shouldn't be the way to solve it. I think there could be a better way to solve this by passing the new date. I am thinking about creating another method which will be something like `updateProps` or something that will pass the new props to the `CustomMarler` implementation. what do you think?
As for your idea, I think this way is okay but I don't see any extra benefits than changing the key which will basically resubscribe a marker.

kevinmanncito:
That was actually my initial thought but I couldn't see a way of getting to the `_id` attribute from the public `CustomMarker` to pass into the `TimelineMarkersProvider ` in order to know which marker to update. Maybe I can use something else to know which marker to update or maybe I was missing something. I can double check again tonight and see if there is a clean way of using an `updateProps` method and possibly update this PR if you'd like.

Ilaiwi:
what about getting another method in the return value from `subscribe` beside `unsubscribe` something like `updateProps`? it would be something like [`rerender`](https://github.com/kentcdodds/react-testing-library#rerender)

Ilaiwi:
and I would really appreciate any help I could get 👍

kevinmanncito:
Okay cool, I'll take a look later today at doing that

kevinmanncito:
@Ilaiwi  updated PR. Let me know if that is what you were thinking.

kevinmanncito:
Sure, I will get to that this week and update the PR

kevinmanncito:
@Ilaiwi  Sounds good, thanks for this library, we are putting it to good use on a project right now. I'll definitely let you know if I run into any more issues or find anything else I could contribute.

Ilaiwi:
Please feel free to do so. You can find issues to contribut to using the help wanted label

Issue #414, opened 2018-09-04
Timeline `GroupRow` height doesn't match `lineHeight` prop


Ilaiwi:
**Describe the bug**
If `lineHeight` prop is set then the `GroupRow` will have a height `lineHeight - 1`. This causes problems with how items look when `itemHeightRatio` is set to `1`.

**To Reproduce**

https://codesandbox.io/s/j3o7q8zpj9

**Expected behavior**
`GroupRow` should have the same height as  `lineHeight` prop

**Screenshots**

![screenshot](https://user-images.githubusercontent.com/10817089/45019536-427c5580-b035-11e8-946c-ba9834d0dc39.PNG)

**Library Version**
0.19

**Desktop (please complete the following information):**

* OS: Windows
* Browser: chrome
* Version: 68.0.3440.106

Ilaiwi:
released in 0.26.3

PR #433, opened 2018-09-28
fix wrong group height

Ilaiwi:
**Issue Number**
#414 

**Overview of PR**
fix `lineHeight` prop not matching the height


Ilaiwi:
this is causing more problems apparently 
I will close this and try to figure out a better solution
screen shot 2018-09-28 at 11 45 54 pm

PR #439, opened 2018-10-05
Allow click at sub header to zoom to second if min zoom is set

tangverapong:
**JIRA: PLAS-3092**

**Overview of PR**
as a developer, i want sub header to be clickable to zoom in to second unit 
and also show the date format at the main header as 'dddd, LL, HH:mm:00' format.


tangverapong:
c

Issue #440, opened 2018-10-07
Resizing the sidebar does not cause update


acemac:
**Describe the bug**
If the sidebar is resized, the component will not update as expected. This is on develop/master 0.20

**To Reproduce**

Resize the sidebar. Can use one of the demos, set the sidebar to the state, and call a setState changing the sidebar's width

goldenjayr:
how do you resize the sidebar?

developer-ocansey:
Closing as part of issue triage. This issue was created before 2023 and has had no recent activity. If you're still experiencing this on v0.30+, please open a new issue with a minimal reproduction.

PR #441, opened 2018-10-07
Fix for resizing the sidebar

acemac:
**Issue Number**

#440 

**Overview of PR**

This is a bug I introduced when trying to increase performance of the scrolling. I missed a case of calculating the new width of the scroll component if the sidebar changes width

### DIFF
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -387,6 +387,9 @@ export default class ReactCalendarTimeline extends Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
+    if (this.props.sidebarWidth !== prevProps.sidebarWidth) {
+      this.resize(this.props)
+    }
     const newZoom = this.state.visibleTimeEnd - this.state.visibleTimeStart
     const oldZoom = prevState.visibleTimeEnd - prevState.visibleTimeStart
     


acemac:
This is a bandaid fix to prevent an ugly bug.

If the sidebar is resized, it will create a second setState to be called immediately after a render has completed.

This could probably be streamlined.

Ilaiwi:
This should all be unit tested 😔. What would a streamlined solution be?

acemac:
I revisited this today. It seems to work... I honestly cannot tell why. I was mid way through changing it and it started working, I've been unable to break the examples so far.

Previously, it was waiting for after `componentDidUpdate` and the render to check if the sidebar's had changed width, then doing a `setState` to cause a re-render with the sidebars at the correct width.

It now uses `getDerivedStateFromProps` to determine if the canvas should be re-rendered.

I will have another look at this soon, and fix the broken tests.

dovydaskukalis:
Hey, any update on this?

Ilaiwi:
@dovydaskukalis I will merge this and #424 during the next 2 weeks

vegimite:
Hi all, any chance that this pull can be resolved in the near future?

Issue #442, opened 2018-10-08
InfoLabel shows [object Object] in tree group when dragging items


thiagosatoshi:
**Describe the bug**

An InfoLabel is displayed when an item is being dragged. When using tree groups, the InfoLabel shows [object Object].

The reason is because the Sidebar and the InfoLabel both use the groupTitleKey to render themselves, but the groupTitleKey can point to a React element for the Sidebar. This is used in the tree group functionality, to simulate collapsing/expanding of the group in the side bar.

**To Reproduce**

Run the demos in the project.
Go to the tree groups demo (https://codesandbox.io/s/k9qyx9zp2o)
Move one of the items.
Observe that the tooltip in the bottom left corner ends in [object Object]

**Expected behavior**

The tooltip should show a suitable text string, for example the title of the group.

**Screenshots**

![timeline bug](https://user-images.githubusercontent.com/2606961/46619323-bb316e80-cb19-11e8-950e-01f0295ec384.png)

**Library Version**

0.20.0

**Desktop (please complete the following information):**

* OS: Windows 10
* Browser: Chrome
* Version: 69.0.3497.100

**Additional context**

The fix involves introducing a new key in the Timeline keys object called `groupLabelKey`, which is defaulted to `title`. This will preserve existing functionality. When the `groupLabelKey` is specified to point to a string attribute of the group object, that is used to render the text in the InfoLabel.

This allows the developer to split what is used to render the Sidebar with what is used to render the InfoLabel.

Ilaiwi:
released in 0.22.0

PR #443, opened 2018-10-08
Added group label key to fix label info onDrag

thiagosatoshi:
**Issue Number**

https://github.com/namespace-ee/react-calendar-timeline/issues/442#issue-367851639

**Overview of PR**

The fix involves introducing a new key in the Timeline keys object called `groupLabelKey`, which is defaulted to `title`. This will preserve existing functionality. When the `groupLabelKey` is specified to point to a string attribute of the group object, that is used to render the text in the InfoLabel.

This allows the developer to split what is used to render the Sidebar with what is used to render the InfoLabel.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/namespace-ee/react-calendar-timeline",
   "repository": {
     "type": "git",
-    "url": "https://github.com/sky-uk/react-calendar-timeline.git"
+    "url": "https://github.com/namespace-ee/react-calendar-timeline.git"
   },
   "author": "Marius Andra <marius.andra@gmail.com>",
   "contributors": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "react-calendar-timeline",
-  "version": "0.20.1",
+  "version": "0.21.0",
   "description": "react calendar timeline",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-calendar-timeline",
-  "version": "0.20.0",
```
Ilaiwi: this is a new feature so it would be 0.21.0

package.json
```diff
@@ -20,7 +20,7 @@
   "homepage": "https://github.com/namespace-ee/react-calendar-timeline",
   "repository": {
     "type": "git",
-    "url": "https://github.com/namespace-ee/react-calendar-timeline.git"
```
Ilaiwi: could you please revert this?

Ilaiwi:
thank you for this. small changes need to be done but everything else is fine

thiagosatoshi:
@Ilaiwi thanks for the review, I have now updated the PR as requested. Please have a look again once you have time. 👍

thiagosatoshi:
@Ilaiwi done. Please let me know if that is OK now?

Ilaiwi:
@thiagosatoshi this will be released in 0.22.0 during the next two weeks thank you for your work

PR #447, opened 2018-10-15
use 'rct' prefix for class names of Header children

Simek:
**Overview of PR**

For most of CSS classes prefix `rct` is used to avoid styling issues. Two children of Header `top-header` and `top-header` currently are missing this prefix which this PR adds.


Ilaiwi:
Thank you for this

PR #448, opened 2018-10-15
Remove duplicated CSS property in 'rct-item-content'

Simek:
**Overview of PR**

Just a small cleanup during work on my implementation, nothing fancy.


Ilaiwi:
good catch thanks!

PR #449, opened 2018-10-16
Fix documentation for verticalLineClassNamesForTime

holytoastr:
**Issue Number**

#452 

The documentation for verticalLineClassNamesForTime as previously written includes a code sample that returns only the first holiday in the list of holidays.

**Overview of PR**

After this change, the sample code in the documentation will return all holidays in the range, per the description for this function.

### DIFF
--- a/README.md
+++ b/README.md
@@ -607,15 +607,18 @@ An example could look like (see: demo/vertical-classes):
 verticalLineClassNamesForTime = (timeStart, timeEnd) => {
   const currentTimeStart = moment(timeStart)
   const currentTimeEnd = moment(timeEnd)
+  let holidayArray = []
 
   for (let holiday of holidays) {
     if (
       holiday.isSame(currentTimeStart, 'day') &&
       holiday.isSame(currentTimeEnd, 'day')
     ) {
-      return ['holiday']
+      holidayArray.push('holiday')
     }
   }
+
+  return holidayArray
 }
 ```
 


Ilaiwi:
Hi @holytoastr, thank you for the PR but could you please open an issue with the problem? I am having a hard time understanding this change

holytoastr:
@Ilaiwi Thanks for the feedback! I opened #452 to better explain myself. Let me know if you have any further questions!

dkarnutsch:
As I've developed the feature and also written the documentation I kind of understand the problem. Namely the example does not make it possible to assign multiple classes, right?

But your fix does not make any sense. You are just assigning the same class (`holiday`) multiple times. One possible solution could look like this: Checking for multiple types (e.g. holiday and sickness or whatever) and then returning the array. But I'm not sure if that is required.

holytoastr:
@dkarnutsch 

> As I've developed the feature and also written the documentation I kind of understand the problem. Namely the example does not make it possible to assign multiple classes, right?

Perhaps I am being unclear. It has nothing to do with multiple classes.

The way the sample code is originally written, it will apply a class name to only the first vertical column that would be included in the array of holidays. My fix allows the code to apply the desired class name to all columns that are in the specified date range.

For example, if your company has both Thursday and Friday off for Thanksgiving, you'd want both date columns to display as holidays. My fix in the sample code is what was needed to display both days of that week as holidays. Using the code as originally written in the documentation resulted in only the Thursday column of that week displaying a holiday.

Like I said on the issue, the library itself works great. It was only a small bit of documentation that caused my stumble and I was hoping to smooth the path for future devs using this code.

dkarnutsch:
@holytoastr 
Please have again a look at the code you are committing.

`verticalLineClassNamesForTime` is returning an array of strings of classNames which are applied to the column. With `holidayArray.push('holiday')` you are adding the className `holiday`. With the return outside of the loop, you could possibly return an array with multiple entries of `'holiday'`.

The function `verticalClassNamesForTime` is called for every column. This should ensure that multiple holidays per month are possible.

dkarnutsch:
@Ilaiwi I think this can be closed for now.

Issue #455, opened 2018-10-24
When zooming in/out events are moved across the timeline


jarnovanrhijn:
**Describe the bug**
When zooming in/out events are moved across the timeline and then do not match the start and end time.

**To Reproduce**

1. Go to https://codesandbox.io/s/zr5r289rm4
2. Zoom in / out
3. See events being moved to the left / right

![screen recording 2018-10-24 at 10 01 32 1](https://user-images.githubusercontent.com/2767127/47415731-755ce300-d774-11e8-95f0-9254ab592723.gif)

**Expected behavior**
Events should match the right start- and endtime when zooming in or out.

**Library Version**
React: 16.0.0
RCT: 0.20.0

**Desktop (please complete the following information):**

* OS: MacOS
* Browser Chrome
* Version 69.0.3497.100

Ilaiwi:
@acemac I have the same issue using the sandbox provided by you but it appears when you zoom in and out with the subheader has day units shown (sorry I do not have gif capture on my osx machine)

acemac:
@Ilaiwi Thats no good! I'll get an MacOS machine to test against.

Every 'tick' of my mousewheel triples the amount of timeline shown. And zooming in from any level immediately zooms to a 1 hour visible range. This has always been the behaviour we see on Win10, Win8 and Windows Server.

The video provided there seems to be far more control over the zoom level. Perhaps a race condition with multiple setState called rapidly on the zoom? [Similar to this issue](https://stackoverflow.com/questions/30338577/reactjs-concurrent-setstate-race-condition)

Ilaiwi:
@acemac yes scrolling events seem to have a different behavior in osx than windows. I don't think it is a race condition though because we don't pass state from children to parents but maybe I don't know much about the race condition problem in react much. 

I will debug this after work and comment my findings here

acemac:
@Ilaiwi Good job :) Is this broken? [Example using setState from parent](https://codesandbox.io/s/l52j57yrz9)

Timeline.js line 495

```
    this.setState(
      calculateScrollCanvas(
        visibleTimeStart, 
        visibleTimeEnd, 
        forceUpdateDimensions, 
        items, 
        groups, 
        this.props, 
        this.state))
  }
```

This is a potential race condition if I understand React correctly.

```
    this.setState(prevState => 
      calculateScrollCanvas(
        visibleTimeStart, 
        visibleTimeEnd, 
        forceUpdateDimensions, 
        items, 
[...]

PR #458, opened 2018-10-25
Fix to force re-render on zoom change

acemac:
**Issue Number**

#455 

**Overview of PR**

Forces canvas to re-draw when the zoom level changes.
On MacOS the zoom level changes a small amount on zoom. This causes the timeline items to not update and appear to 'drift' on the timeline.


Ilaiwi:
@acemac thank you for the PR, this fix actually didn't work so I had to revert it. I am doing another PR with unit testing. I will tag you on it for review if you would like?

PR #391, opened 2018-07-30
Custom headers

Ilaiwi:
**Issue Number**

#349 

**Overview of PR**

this PR adds the following features:

1. render more headers in the header section 
2. add the ability to add completely customized headers.

To checkout, the work run the demo and chose `customHeaders` demo

customheaders

 _**This work is still really really early and only to show core concept of the feature and to show progress**_

### TODO

- [x] fix drag items bug where extra custom headers cause items to not be correctly positioned correctly dragging
- [x] convert to propGetter pattern
- [x] add style and classnames props to customize headers root element

- [x] add props to style root element and add classnames to root element

- [x] add render prop for `DateHeader` for intervals
- [x] fix broken tests

- [x] add documentation

- [x] add tests for custom headers

- [x] delete format props
- [x] refactor and test 
- [x] decide what to do with sticky header prop
- [x] handle `headerRef` prop

- [x] maybe unify `CustomHeader` and `VerticalLines` label width calculation

- [x] make sure plugins don't break

- [x] delete sidebarContent and rightSidebarContent readme

- [x] delete `headerLabelGroupHeight` and `headerLabelHeight`
- [x] delete header css classes
- [x] custom header/day header pass primary/secondary via unit
- [x] pass components and functions for render prop
- [x] remove format object option from labelFormat
- [x] default styles in DateHeader => move to classnames and don't provide them if provide interval renderer
- [x] fix custom header docs (interval)
- [x] must pass height to custom headers? how do we fix this?
- [x] check if programmatically scrolling example work
- [x] write breaking change docs
- [x] cleanup old code
- [ ] handle touch state in `TimelineElementsHeader`

- [ ] fix a bug, where headers interval borders are not aligned with the vertical lines, form the calendar

- [ ] add an option to deal with `unit` as an interval of time
- [ ] pass `unit` as interval iterateTimes

PR #472, opened 2018-11-21
Wrap timeline in flex container

richard-still:
Firstly, I'm new to this so apologies if I'm not doing this right!

**Overview of PR**

Just wrapping the timeline in a flex container in the same way that the header is wrapped in a flex container. 

I think there are sometimes issues with the way the width of the viewport is calculated, which in some circumstances can make the timeline go onto two rows like the screenshot below. Resizing the screen forces a recalculation and it renders properly. You can also see this in the codesandbox examples if you slowly resize the preview pane to be smaller, the items jump down (to be in the same position as this screenshot) then they re-render to be in the right place. 

Wrapping the sidebar and timeline in a flex container allows the siderbar to shrink, in the same way that the header sidebar shrinks, and always keeps the timeline on the same row

image

### DIFF
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -1006,64 +1006,67 @@ export default class ReactCalendarTimeline extends Component {
               headerLabelGroupHeight,
               headerLabelHeight
             )}
-            {sidebarWidth > 0 && this.sidebar(height, groupHeights, headerHeight)}
-            <div style={{display: 'inline-block'}}>
 
-              <div style={outerComponentStyle} className="rct-outer">
-              <ScrollElement
-              scrollRef={el => {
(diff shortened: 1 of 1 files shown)


Ilaiwi:
Hi, Thank you for this. I really appreciate your help. Could you please open an issue with the bug so we can connect this to the issue for future documentation.

This will be released in the next feature version that will go out this weekend

richard-still:
Thanks @Ilaiwi. Much appreciated

Ilaiwi:
this will be solved when #391 gets merged. Thanks for reporting. Please refer to #476 for more details

Ilaiwi:
this will be solved when #391 gets merged. Thanks for reporting. Please refer to #476 for more details

Issue #350, opened 2018-06-28
onItemDeselected function is not documented


yftachman:
**Describe the bug**
Hey, I've found a function on the props that is not documented and works. Also makes sense to have this function (it was just what I needed).

Weird part is it has a comment on it saying this isn't in the docs...

mcMickJuice:
Seems like a quick doc fix. Mind submitting a PR?

Ilaiwi:
released in 0.26.3

PR #499, opened 2019-01-23
fix selected prop

Ilaiwi:
**Issue Number**

#350 

**Overview of PR**

- fix controlled vs uncontrolled select.

- [x] add documentation for `onDeselect`


tobiasbueschel:
Great work @Ilaiwi and exactly the fix that I need for an error that we are facing with `react-calendar-timeline`. When I selected an item, then filter the timeline by items and the item is no longer visible, then clicking on the canvas results in the following:

> TypeError: Cannot read property '__interact_scope_95' of null

Also related to what I described here: https://github.com/namespace-ee/react-calendar-timeline/issues/573

Is there anything that's blocking this PR from going in? Could we possibly tackle the outstanding item in a subsequent PR? 
> * [ ]  add documentation for `onDeselect`

Thanks!

src/lib/items/Item.js
```diff
@@ -414,7 +414,6 @@ export default class Item extends Component {
 
   componentDidUpdate(prevProps) {
```
tobiasbueschel: Out of curiosity and because I'm not yet very familiar with the internal codebase of `react-calendar-timeline`. Should we not call `componentWillUnmount` to unsubscribe from any `interact` event listeners on the `item`?

Ilaiwi: I dont think interactjs has an unmount or cleanup call

Ilaiwi:
@tobiasbueschel could you help to add docs please

src/lib/items/Item.js
maartenpeels: Looking at the [docs](https://interactjs.io/docs/faq/#remove-%2F-destroy-%2F-release), it looks like there is a method: `interact(target).unset();`

Ilaiwi: interesting. I will open an issue to look into that

Ilaiwi: I actually can't find the documentation for `unset` could you please link it here?

PR #509, opened 2019-02-13
docs: correct invalid JavaScript syntax in examples

tobiasbueschel:
**Issue Number**

This PR does not relate to any issue as it's merely a documentation fix.

**Overview of PR**

Fixes invalid JavaScript syntax, i.a. removes explicit return statements from [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions#Function_body) with concise body as show in the examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -711,10 +711,10 @@ const twoSeconds = 2000
 
 <TodayMarker>
   {({ styles, date }) =>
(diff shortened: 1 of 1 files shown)


Ilaiwi:
@tobiasbueschel  thank you for this.

PR #510, opened 2019-02-14
feat(Timeline): Add UTC offset support

MadDeveloper:
**Issue Number**
https://github.com/namespace-ee/react-calendar-timeline/issues/512

**Overview of PR**

This PR will add an UTC offset support to the whole calendar timeline.

`utcOffset` and `applyLocalOffset` prop have been added to the library.
They permit to set an offset from UTC (`utcOffset`), and to keep some informations (markers) correlated to the local timezone offset (`applyLocalOffset`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,10 @@ The exact viewport of the calendar. When these are specified, scrolling in the c
 
 **Note that you need to provide either `defaultTimeStart/End` or `visibleTimeStart/End` for the timeline to function**
 
+## utcOffset
+
+Define the offset you want to apply from UTC. Defaults value is your local timezone offset (e.g. moment().utcOffset())
+
 ## selected
 
 An array with id's corresponding to id's in items (`item.id`). If this prop is set you have to manage the selected items yourself within the `onItemSelect` handler to update the property with new id's. This overwrites the default behaviour of selecting one item on click.
@@ -696,6 +700,10 @@ Marker that is placed on the current date/time.
 > interval: number | default: 10000
 
 How often the TodayMarker refreshes. Value represents milliseconds.
+
+> applyLocalOffset: boolean | default: true
+
+If you want to apply the local timezone offset when you display some informations (e.g. the today marker). This option is often correlated with `utcOffset` option, and should be provided to the `<Timeline>` component.
 
 > children: function({styles: object, date: number}) => JSX.Element
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,4 @@
+
 # Change Log
 
 All notable changes to this project will be documented in this file.
@@ -6,6 +7,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres (more or less) to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
+
+### 0.24.0
+
+### Added
+
+* utc offset support
 
 ## 0.23.0
 
--- a/__tests__/components/Header/Header.test.js
+++ b/__tests__/components/Header/Header.test.js
@@ -22,6 +22,7 @@ const defaultProps = {
   stickyHeader: true,
   headerLabelGroupHeight: 15,
   headerLabelHeight: 15,
+  utcOffset: 0,
   scrollHeaderRef: () => {},
   headerRef: () => {}
 }
--- a/__tests__/components/Header/TimelineElementsHeader.test.js
+++ b/__tests__/components/Header/TimelineElementsHeader.test.js
@@ -16,6 +16,7 @@ const defaultProps = {
   subHeaderLabelFormats: {},
   headerLabelGroupHeight: 0,
   headerLabelHeight: 0,
+  utcOffset: 0,
   scrollHeaderRef: () => {}
 }
 
--- a/__tests__/components/Markers/CustomMarker.test.js
+++ b/__tests__/components/Markers/CustomMarker.test.js
@@ -67,7 +67,9 @@ describe('CustomMarker', () => {
       visibleTimeEnd,
       canvasTimeStart: visibleTimeStart - oneDay,
       canvasTimeEnd: visibleTimeEnd + oneDay,
-      canvasWidth
+      canvasWidth,
+      utcOffset: 0,
+      applyLocalOffset: false
     }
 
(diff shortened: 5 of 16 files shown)


tobiasbueschel:
@MadDeveloper thanks a lot for this PR - exactly what I was looking for as well :-)

@Ilaiwi & @mcMickJuice is there anything that is holding this PR from being merged besides the merge conflicts?

README.md
```diff
@@ -130,6 +130,10 @@ The exact viewport of the calendar. When these are specified, scrolling in the c
 
 **Note that you need to provide either `defaultTimeStart/End` or `visibleTimeStart/End` for the timeline to function**
 
+## utcOffset
+
+Define the offset you want to apply from UTC. Defaults value is your local timezone offset (e.g. moment().utcOffset())
```
tobiasbueschel: What input does `utcOffset` take?

MadDeveloper: String or number.
[https://momentjscom.readthedocs.io/en/latest/moment/03-manipulating/09-utc-offset/](https://momentjscom.readthedocs.io/en/latest/moment/03-manipulating/09-utc-offset/)

tobiasbueschel: Cool 👌 perhaps we can add this as an example input to the `README.md`? What do you think?

tobiasbueschel:
> @MadDeveloper thanks a lot for this PR - exactly what I was looking for as well :-)
> 
> @Ilaiwi & @mcMickJuice is there anything that is holding this PR from being merged besides the merge conflicts?

@MadDeveloper would you have a chance to update the merge conflicts? 🙏 I'm super keen to get this in sooner than later!

MadDeveloper:
@tobiasbueschel I would like to see this PR merged, and if I would know that it's gonna be done, I will spend time fixing conflicts, but this PR is opened since a certain time and nobody is giving interest... So I created my own fork, with my updates, you can use it if you want: [https://github.com/MadDeveloper/react-calendar-timeline](https://github.com/MadDeveloper/react-calendar-timeline)

tobiasbueschel:
> @tobiasbueschel I would like to see this PR merged, and if I would know that it's gonna be done, I will spend time fixing conflicts, but this PR is opened since a certain time and nobody is giving interest... So I created my own fork, with my updates, you can use it if you want: https://github.com/MadDeveloper/react-calendar-timeline

We're in the same boat and also run our own fork at the moment! :-)

@Ilaiwi & @mcMickJuice sorry for the additional ping, but is there a chance that @MadDeveloper and I can help with getting this merged in? Or perhaps there is any other way we can help with the ongoing efforts to improve `react-calendar-timeline`?

README.md
```diff
@@ -130,6 +130,10 @@ The exact viewport of the calendar. When these are specified, scrolling in the c
 
 **Note that you need to provide either `defaultTimeStart/End` or `visibleTimeStart/End` for the timeline to function**
 
+## utcOffset
```
tobiasbueschel: If we apply a `utcOffset` to the Timeline how does this PR handle timeline events such as `onItemMove` and `onItemResize`?

Will these callbacks send back UTC timestamps or timestamps with the UTC offset still applied?

MadDeveloper: I didn't use it since this PR, so what I will say need to be verified : if you give an utcOffset, all computation are relative to the offset, so normally, you will receive in callback, times with the offset applied.

tobiasbueschel: I'm thinking that we might want to convert it back to UTC in the callback. The reason is the following scenario:

- Database stores timestamps in UTC
- API sends back data in UTC
- Timeline uses UTC offset to show these items in the user's timezone
- When a user updates the time of the item, the callback is used to send back the timestamp in UTC so that the API can store it in the DB

In this way, all data that goes **into** the timeline is UTC and all data that goes **out** of the timeline is in UTC. Or if you're not using an `utcOffset` all data going into the `timeline` comes out in the same way.

What do you think?

MadDeveloper: I think `react-calendar-timeline` doesn't have to take care of that, it outputs what you asked him, and you still able to convert the time in utc in callbacks 😉

tobiasbueschel: Hmm, but then you could argue that you might as well just convert your input `items[n].start_time` and `items[n].end_time`, `canvasTimeStart`, etc. to the UTC offset that you need, right?

@Ilaiwi curious to hear your thoughts as I'm pretty new to the codebase of `react-calendar-timeline`

MadDeveloper: I think especially that `react-calendar-timeline` is not a time converter, instead of `moment`

Ilaiwi:
@tobiasbueschel @MadDeveloper So I've been thinking about this one for a while not and I think the correct way to go forward is to implement something similar to what https://material-ui-pickers.dev/ does via @date-io.
We would basically inject moment in this case via the react api with the offset and locals appropriate set in the moment instance. This will give more control to the user and help us support other date utils libs....
https://github.com/dmtrKovalenko/date-io

I am very interested in knowing what do you guys think...

spiroski:
Any progress on this issue? I believe displaying the timeline in a specific timezone is a feature that many users of this library could benefit from.

MadDeveloper:
@Ilaiwi indeed `date-io` can be used, but it's more related to a refactor than a feature. So I think this PR can be firstly added, and then refactor the core library with `date-io`.

Ilaiwi:
@MadDeveloper hmmm yes I agree. However, I think this feature should not be a part of this library but be inverted to the user in. This is possible via date-io. I see your point though.
Could you please solve the conflict. I want to take another look at this... maybe as you said we can merge this and refactor later...

MadDeveloper:
@Ilaiwi I tried to merge but it seems complicated, the codebase has meaningful changed. I'll try later, but if it's too much complicated, maybe try to implement directly `date-io`, or apply me changes directly from a new branch checked out from master.

Ilaiwi:
@MadDeveloper I would lean towards doing the `date-io`. After thinking about this I would rather not support something than support it then remove it in the future

MadDeveloper:
@Ilaiwi I agree, that's a better philosophy. Will avoid futur issues due to this intermediate solution.

Ilaiwi:
@MadDeveloper Thanks for understanding.

Ilaiwi:
I will close this since it doesn't match the library's interest as explained in the comments above

Issue #415, opened 2018-09-05
Commit e090aeec appears to break item styling


mrampersad:
**Describe the bug**
As of commit e090aeec, item class override for overridable styles doesn't work.

**To Reproduce**

_Please provide one or both of the following:_

1. Try to override the background color of an item using a class name according to the documentation.

https://github.com/namespace-ee/react-calendar-timeline#how-can-i-have-items-with-different-colors

**Expected behavior**
The background color is set according to the class name.

**Screenshots**
![image](https://user-images.githubusercontent.com/451351/45119472-910b1c00-b129-11e8-923a-207f42476af9.png)

**Library Version**
0.19.0

**Desktop (please complete the following information):**

* OS: Windows 10 x64
* Browser: Chrome
* Version: 68.0.3440.106

**Additional context**
I believe that the introduction of src/lib/items/styles.js moves the background color from class to style directly on the element.

Ilaiwi:
Well, to be honest, I thought about that. You can, of course, override the inline styles using  [getItemProps](https://github.com/namespace-ee/react-calendar-timeline#itemrenderer), but I think this depends on the told you are using to style your components. 
A solution I thought of is to move all the unnecessary styles to classes but if you are using `itemRenderer` then you will not get these classes in the return value for `getItemProps`. What do you think?

mrampersad:
itemRenderer is an advanced interface that I am not using. I cannot comment on how the solution will impact users of itemRenderer. It sounds like this will restore the ability to customize the look of items using class names, so I like it.

mcMickJuice:
@mrampersad Why don't you just use itemRenderer? it's purpose is use cases like this.

mrampersad:
@mcMickJuice The reason I have not considered itemRenderer is that according to the FAQ, changing the look of an item is two lines of effort: add a className and define it in CSS. Either the new behavior should be considered a bug, or the FAQ should be updated to put in a working example of changing the color. itemRenderer looks very powerful, but it is dozens of lines of reading with concepts that I haven't needed to understand thus far.

The method described in the FAQ is a reasonable middle-ground between zero customization and implementing itemRenderer myself. After reviewing the itemRenderer documentation for a few minutes, it seems like instead of a two-line change, I might have to add 40 lines to change the colors. Is that really what you want developers to do?

[...]

PR #516, opened 2019-02-21
Fix/item style prop

mamihelj09:
**Issue Number**
#415 

**Overview of PR**

This PR will fix style props in ```getItemProps()``` method

Props in  ```getItemProps()``` method is an object containing title of the item. 
Like getting the className from props on line 514 i used the same logic to get the styes from props.


Ilaiwi:
@mamihelj09 Thank you for the PR but I think you misunderstood the problem. The problem is now where from where do we get the props but what do we provide with the default implementation.

If you have any questions please ask them in the issue  and I will make sure to answer them asap.

Issue #517, opened 2019-02-22
Ability to customize the InfoLabel


FutureArchitect-wang2449:
**Is your feature request related to a problem? Please describe.**
It could be nice if we show customized information while dragging (moving or resizing) items (currently only showing time and groupLabelKey).

**Describe the solution you'd like**
Provide a renderer for InfoLabel the same way like groupRenderer or itemRenderer.

**Describe alternatives you've considered**
At least give an option to hide InfoLabel while dragging. (Sometimes you do not like it to show)

**Additional context**
I checked your feedback of the request to customize the header (#48)
But I think this one is much easier and really does not take so much resource, so I will be very graceful this time if you do not close this one that recklessly.

FutureArchitect-wang2449:
Finally I added the [simple function](https://github.com/namespace-ee/react-calendar-timeline/issues/517#ref-pullrequest-413946637) by myself (a personal account).
I will be grateful if you have time to review and pull it. (This will be the first time I contribute to oss, excited! )

@mariusandra @Ilaiwi

Ilaiwi:
@FutureArchitect-wang2449 Thank you for your first PR. I really appreciate your help. I have another suggestion that will be more helpful. Let's add a new prop called `onDragUpdate` and remove the infoLabel component. If a user wants to add an InfoLabel they can do it via this callback Prop.

It is also a good idea to include a sandbox example on that. it would be also great if you would provide an example of an added on `infoLabel` with his new prop.

The code would look something like this: https://github.com/FoothillSolutions/react-calendar-timeline/commit/a43f8e5cb89a3a0552e426d3590777df86ed9b4e

Good luck :)

FutureArchitect-wang2449:
@Ilaiwi 
> Let's add a new prop called onDragUpdate and remove the infoLabel component.

Thank you for your suggestion. I totally aggree your points of view.

However, I noticed there is already one prop which is triggered when items are being dragged. That prop is `moveResizeValidator(action, itemId, time, resizeEdge)`, although it is not for render something, instead, for validation (as its name describes).

Further more, there is also a `groupLabelKey` in the `keys` prop which is all the way for this `InfoLabel`, so I suppose the developers of this component laid strong emphasis to the `InforLabel`

Considering the present situation, I added the renderer for the `InfoLabel` instead of a total free handler.

I appreciate if you take the current work (that they have done for `InfoLabel`) in consideration.

Ilaiwi:
[...]

PR #518, opened 2019-02-25
add props infoLabelRenderer

wsysuper:
**Issue Number**

#517

**Overview of PR**

added props `infoLabelRenderer` to allow user customize info label when dragging an item

(the contents below are also added to README.md)

## infoLabelRenderer(action, itemId, time, newGroupOrderOrResizingEdge)

React component that will be used to render the content of infoLabel while moving or resizing items.

The argument `action` is one of `move` or `resize`.

The argument `time` describes the proposed new time for either the start time of the item (for move) or the start or end time (for resize).

The argument `newGroupOrderOrResizingEdge` is the group order (for move), or one of `left` or `right` (for resize).

The function should return a string or a react component, or `null` for disabling infoLabel.

For example, use this code:

```jsx
infoLabelRenderer = (action, itemId, time, newGroupOrderOrResizingEdge) => {
  let label = null
  if (action == 'move') {
    let newGroup = this.state.groups[newGroupOrderOrResizingEdge]
    const groupLabel = newGroup ? newGroup.title : ''
    label = `${moment(this.state.dragTime).format('LLL')},
        ${groupLabel}`
  } else if (action == 'resize') {
    label = `${moment(time).format('LLL')}, ${newGroupOrderOrResizingEdge}`
  }
  return label
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,11 @@ and this project adheres (more or less) to [Semantic Versioning](http://semver.o
 
 ## 0.23.0
 
-- improve unit tests coverage #426 - @ilaiwi
-- stack items by group #384 - @acemac
-- fix bug where `canMove` prop gets ignored #484 - @acemac + @ilaiwi
-- fix sidebar re-render when groupHeights do not change #478 - @SDupZ
+* improve unit tests coverage #426 - @ilaiwi
+* stack items by group #384 - @acemac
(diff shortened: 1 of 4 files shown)


Ilaiwi:
@wsysuper  Thank you for the PR. Could you please take a look at the issue. I think there is a better approach to solve this.

PR #560, opened 2019-04-19
Fix/mouse wheel passive

### LINKED ISSUE #541: Chrome 73 Treats Document Level Wheel/Mousewheel Event Listeners as Passive
**Describe the bug**

When zoom in using the touchpad, browser gives many duplicated console errors as below.
```
[Intervention] Unable to preventDefault inside passive event listener due to target being treated as passive. See https://www.chromestatus.com/features/6662647093133312
```

**To Reproduce**

Using MacOS Google Chrome Version 73.0.3683.86 (Official Build) (64-bit)
When Using the touch pad to zoom in at then timeline canvas, the error logs come out.
And not only the date range changes, also whole page is zoomed in.

**Expected behavior**
When zoom in guesture at the timeline canvas, only the date range changes, the whole page should not be zoomed in.

**Library Version**
"react-calendar-timeline": "^0.23.0",

**Desktop (please complete the following information):**

* OS MacOS Majove 10.14.3
* Browser Chrome
* Version  >= 73

wsysuper:
@Ilaiwi 

**Issue Number**

Close #541 

**Overview of PR**

fix the Chrome 73 wheel event by adding `passive: false` option.

ref:
https://github.com/facebook/react/issues/14856#issuecomment-478144231


Ilaiwi:
thank you for the effort. I already made a PR to fix it. I will push it to master very soon

Issue #562, opened 2019-04-19
TimeLine crashes for DateHeader with a unit of week 


dkarnutsch:
**Describe the bug**
When using a `DateHeader` with `unit=week` the timeline crashes if no label is provided, because a default label is not set. 

**To Reproduce**
See the following Code Sandbox: https://codesandbox.io/s/64xmvj421w?fontsize=14

**Expected behavior**
Default labels should be applied

**Library Version**
0.25.0

Fixed in #563

Ilaiwi:
good catch!

stahlmanDesign:
Related to DateHeader:
The examples are not up to date with v0.25.0, and the docs seem wrong

@dkarnutsch Your Code Sandbox helped me figure out you have to use `unit='primaryHeader'` above the secondary DateHeader component to get the two levels
``` javascript
<DateHeader unit='primaryHeader' />
<DateHeader />
```

@Ilaiwi The docs show it as a bool:
``` javascript
<DateHeader primaryHeader />
```
but that doesn't work, nor does
``` javascript
<DateHeader primaryHeader={ true } />
```

Ilaiwi:
@stahlmanDesign hmmm I can't find the lines in docs that you are refaring to. could you link them here so I can fix them really quick?

stahlmanDesign:
> @stahlmanDesign hmmm I can't find the lines in docs that you are refaring to. could you link them here so I can fix them really quick?

I'm sorry I can't find it either now. I think I was looking at a bad example. The docs are OK.

Ilaiwi:
@stahlmanDesign that was an old version of the documentation what you were referring to. I would appreciate it if you can create a good custom headers example and add it to docs

Ilaiwi:
released in 0.25.1

PR #563, opened 2019-04-19
add default label formats for week

dkarnutsch:
**Issue Number**

#562


tobiasbueschel:
Nice fix @dkarnutsch, is there any value in adding some tests so that we can ensure this crash won't happen again?

PR #586, opened 2019-06-11
Undefined classnames in TimelineHeaders

trevdor:
**Issue Number**

566

**Overview of PR**

Use [`classnames` utility](https://github.com/JedWatson/classnames) to build lists of classnames, automatically excluding falsey values.

Result:
no_undefined_classnames

Commits were separated for clarity but are intended to be squashed into one.

**Alternative Approach**
If the project does not want the new dependency on `classnames`, we could achieve the same exclusion of falsey values using just native JS, something like:
```jsx
const classnames = (...names) => names.filter(Boolean).join(" ");
className={classnames('rct-header-root', this.props.className)}

```
(Actually, after I write that out, I may like it better since the `classnames` package does a lot this project probably doesn't need. Happy to change it.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres (more or less) to [Semantic Versioning](http://semver.o
 
 ## Unreleased
 
+## 0.25.3
+
+- Fixed the `undefined` classnames in TimelineHeaders #566 @trevdor
+
 ## 0.25.2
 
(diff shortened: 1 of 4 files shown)


Ilaiwi:
@trevdor thank you for the great PR. I really appreciate the unit tests you added and adding the change log. This will make releasing this a breeze

Issue #598, opened 2019-06-20
Broken codesandbox examples


trevdor:
Codesandbox examples are broken that use react-calendar-timeline@0.25.3 (or "latest"), including those on the examples page:
https://github.com/namespace-ee/react-calendar-timeline/tree/master/examples

A couple examples:
https://codesandbox.io/s/ly3oj7xxwm
https://codesandbox.io/s/w6xvqzno4w

### Root Cause
In my v0.25.3 PR, I put the new `classnames` dependency in devDependencies rather than regular production dependencies. It appears [CodeSandbox does not install devDependencies](https://github.com/codesandbox/codesandbox-client/issues/924).

trevdor:
If my analysis of the root cause is correct, the above pull request should fix it. 
Either way, I'm sorry for the trouble.

Ilaiwi:
@trevdor thank you for the help. your PR fixed the issue.

PR #599, opened 2019-06-20
fix: classnames should be a prod dependency, not dev

trevdor:
Issue #598 

**Overview of PR**

Moves `classnames` to a production dependency


Ilaiwi:
@trevdor thanks for the help

Issue #473, opened 2018-11-24
StackElements does not work with Date Objects


sm1ddy:
I've noticed that StackItems does not work when trying to use Date Objects on items. On the other hand, using moment seems to work.

Ilaiwi:
you can either use moment or UNIX milliseconds only we dont support native date

tyson-kubota:
I also ran into this issue; it should be mentioned in the documentation for clarity. I'll open a PR.

PR #661, opened 2019-09-30
Clarify stackItems option timestamp requirements

tyson-kubota:
**Issue Number**

https://github.com/namespace-ee/react-calendar-timeline/issues/473

**Overview of PR**

The `stackItems` option does nothing if the timeline objects use native JS Date objects. Instead, they must be milliseconds or Moment timestamps.

This PR updates the documentation to reflect the above.


Ilaiwi:
thank you

PR #683, opened 2019-11-13
Add className prop to top-level Timeline

### LINKED ISSUE #682: Allow custom className for root element
**Is your feature request related to a problem? Please describe.**

The root element has the `react-calendar-timeline` class name, which can't be customised. (There's also the `rct-outer` class name underneath it as well.)

**Describe the solution you'd like**

I'd like to be able to pass `className` or `additionalClassName` to the top-level `Timeline`.

**Describe alternatives you've considered**

It could be possible to override the class name and change it entirely, rather than just adding a class. I have no particular preference on this.

rmccue:
**Issue Number**

Fixes #682.

**Overview of PR**

Adds a `className` prop to the top-level `Timeline` component to override the `react-calendar-timeline` class.

### DIFF
--- a/src/lib/Timeline.js
+++ b/src/lib/Timeline.js
@@ -209,7 +209,7 @@ export default class ReactCalendarTimeline extends Component {
     itemTouchSendsClick: false,
 
     style: {},
-    className: 'react-calendar-timeline',
+    className: '',
     keys: defaultKeys,
     timeSteps: defaultTimeSteps,
     headerRef: () => {},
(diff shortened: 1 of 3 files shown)


src/lib/Timeline.js
```diff
@@ -208,6 +209,7 @@ export default class ReactCalendarTimeline extends Component {
     itemTouchSendsClick: false,
 
     style: {},
+    className: 'react-calendar-timeline',
```
Ilaiwi: `react-calendar-timeline` should always exist on className I would suggest appending whatever class passed to `react-calendar-timeline`

rmccue: Updated; left the prop name as `className` which matches the other components' props, and updated the docs.

Issue #705, opened 2020-01-02
onCanvasDoubleClick and onCanvasContextMenu documented parameters are wrong


joncar:
**Describe the bug**
The documentation suggests onCanvasDoubleClick and onCanvasContextMenu have a group parameter as an object, which is different from all of the item events which use item ID. Fortunately the code does the right thing and it is actually the group ID.

**To Reproduce**

n/a

Ilaiwi:
released in 0.28.0

PR #706, opened 2020-01-02
Fix docs for onCanvasDoubleClick and onCanvasContextMenu

joncar:
The documentation suggests onCanvasDoubleClick and onCanvasContextMenu have a group parameter as an object, which is different from all of the item events which use item ID. Fortunately the code does the right thing and it is actually the group ID.

**Issue Number**

#705 

**Overview of PR**

Fixes the documented parameters.


Ilaiwi:
thank you for the fix

PR #724, opened 2020-03-05
Add minHeight Props to Timeline to  set the minimum height of the timeline scroll body

markstock7:
Add minHeight to be able to satisfy the following requirements.

屏幕快照 2020-03-04 下午7 44 23


Ilaiwi:
@markstock7 the goal here is to set a min height per group you mean? if so we have a property in the group object called `height` that will do that.

PR #729, opened 2020-03-18
Fix injecting custom class names for vertical lines

RafikiTiki:
Fixes injecting custom class names for vertical lines spanning over a period of time longer than day (eg. month)

Right now if we want to inject custom class name for vertical line spanning month it's class names end up corrupted. If I'll return `timeline-month` class from `verticalLineClassNamesForTime` function I'll end up with vertical line element: `<div class="rct-vltimeline-month">` instead of `<div class="rct-vl timeline-month">`


RafikiTiki:
Quick hacky workaround before this gets merged: make verticalLineClassNamesForTime return class names array with first element as empty string (eg. `return ['', 'timeline-month', 'timeline-month-odd']`

Issue #514, opened 2019-02-20
Include group in moveResizeValidator


chriemmy:
**Is your feature request related to a problem? Please describe.**
I have a set of rules defined in `moveResizeValidator`. They ensure that certain time ranges cannot be allocated. The rules can be defined fine when I have only one group and the validator works.
But when I move an item to a different group I would like to define different time ranges where an item cannot be placed for that particular group.

**Describe the solution you'd like**
Add new parameter to `moveResizeValidator` function **`group`**. It would contain new group for item which is moved by user.

Ilaiwi:
@chriemmy yes this makes perfect sense. This would be a great first PR for anyone who would like to contribute to this library

acemac:
I did this as a part of #412

developer-ocansey:
Closing as part of issue triage. This issue was created before 2023 and has had no recent activity. If you're still experiencing this on v0.30+, please open a new issue with a minimal reproduction.

PR #787, opened 2020-11-02
limit item dragging between groups (improve moveResizeValidator via dragGroupDelta)

Dexmaster:
**Issue Number 514**

The closest issue I could find is https://github.com/namespace-ee/react-calendar-timeline/issues/514
moveResizeValidator for dragging should be able to return time and groupDelta to block drag to specific group.

**Overview of PR**

Two more arguments added to moveResizeValidator:
 - dragOrderIndex (original group index of item)
 - dragGroupDelta (group shift difference on dragging)

Now moveResizeValidator can return either:
 - timestamp in ms (integer) _as before_
 - object with `dragTime` timestamp in ms (integer) and `dragGroupDelta` final group index shift (integer) _new_

Added [small demo](https://github.com/namespace-ee/react-calendar-timeline/pull/787/files#diff-dd31c573e1ac62912d65047e919bfd0df63491c87aaee6550d04612c26bc1874R145-R151) with even/odd movement validation (item from even row can be moved to only even row and vice versa)

### DIFF
--- a/README.md
+++ b/README.md
@@ -316,7 +316,7 @@ Called when the canvas is clicked by the right button of the mouse. Note: If thi
 
 Called when the timeline is zoomed, either via mouse/pinch zoom or clicking header to change timeline units
 
-## moveResizeValidator(action, itemId, time, resizeEdge)
+## moveResizeValidator(action, itemId, time, resizeEdge, dragOrderIndex?, dragGroupDelta?)
 
 This function is called when an item is being moved or resized. It's up to this function to return a new version of `change`, when the proposed move would violate business logic.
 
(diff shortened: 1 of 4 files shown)


arenac (commented):
Is this change covered by tests?

lnagel:
@developer-ocansey — Flagging this for your review.

**Recommendation: Worth porting to the current TypeScript codebase**

### What it does
Extends `moveResizeValidator` to accept two additional arguments (`dragOrderIndex` and `dragGroupDelta`) and return either a number (backward-compatible) or an object `{ dragTime, dragGroupDelta }`, allowing consumers to constrain both time and group during drag operations.

### Why it's valuable
The ability to constrain item dragging to specific groups is a real scheduling requirement (e.g., "items from even rows can only move to even rows"). The current `moveResizeValidator` only returns a time value with no way to control the group. This PR's approach is backward-compatible — if the validator returns a number, it works as before; if it returns an object, both `dragTime` and `dragGroupDelta` are applied.

The implementation is clean:
- New `validateDragState` method extracts and centralizes validation logic
- Type checking handles both return types (`typeof result === 'number'` vs `'object'`)
- The demo clearly illustrates a checker-pattern movement constraint
- A reviewer asked about test coverage — no tests were included, so those would need to be added

### What needs adaptation
The PR targets the old `Item.js`, now `Item.tsx`. The current `moveResizeValidator` TypeScript type is:
```typescript
moveResizeValidator?: (action: 'move' | 'resize', item: TimelineItemBase<any>, time: number, resizeEdge?: ResizeEdge | null) => number
```
This would need to be extended to accept the additional params and support the union return type. The `validateDragState` method should be added to the current `Item.tsx` (the drag handler structure with interact.js is still the same).

### Effort estimate
Medium — focused changes to `Item.tsx` and its type definitions. The interact.js drag handler call sites at lines ~336 and ~357 would need the same enhancement.

developer-ocansey:
Thank you for this contribution! Unfortunately the codebase has been substantially rewritten since this PR was opened (full TypeScript migration, new test suite, and various architectural changes). This PR is now too outdated to merge cleanly. Please feel free to reimplement this feature against the latest `main` branch — the new codebase should make it easier to build on.

Issue #642, opened 2019-08-29
CustomHeader using componentWillReceiveProps


arturcarvalho:
I'm getting the warning about using componentWillReceiveProps that will be renamed. 

![image](https://user-images.githubusercontent.com/2487710/63948297-adaa0d80-ca78-11e9-9fac-0390074834b1.png)

I can fix it, but before sending the I'm checking if the repo is active.

And thanks for the library!

Ilaiwi:
we are active

arturcarvalho:
> 
> 
> we are active

Thank you for your reply.

Are you accepting PRs? I sent one some days back to fix some typos but I had no reply: https://github.com/namespace-ee/react-calendar-timeline/pull/639

Ilaiwi:
@arturcarvalho feel free to submit the PR and I will take a look at it asap.

arturcarvalho:
@Ilaiwi great, thanks! I think I'll be able to submit a PR in the upcoming week.

conorstew:
<img width="1701" alt="Screen Shot 2022-09-06 at 11 32 55 AM" src="https://user-images.githubusercontent.com/71521539/188712403-1c64c63b-cf7d-4dc7-b918-ec2bb0c4d12b.png">

I am still getting this error. Is there a fix for this?

PR #819, opened 2021-05-25
Replace moment with datedriver.

bumpusfrancus:
* Create the dateDriver service.
* Add the vanilla dateDriver with the required moment api.
* Replace all references to moment in the base code with dateDriver.
* Update tests to expect dateDriver.
* Add annotations to tests and test suites failing due to deprecations.

**Issue Number**

Link to issue this PR addresses.  If no issue exists, please provide reasoning for PR.

**Overview of PR**


Ilaiwi:
@bumpusfrancus really great work. Are parts of the implementation of `vanillaDateDriver.js` a derivative version of `moment`?
As for the next step for this, I will use this PR in the v1 version I am drafting. As for merging it with the current master, I will look into it but for it to be merged I would fix the skipped tests and add update the documentation

deleonio:
@bumpusfrancus, @Ilaiwi 

What does the removal of `componentWillReceiveProps` mean.

Are the skipped tests irrelevant in the future?

bumpusfrancus:
Using componentWillReceiveProps causes React to log a warning, and due to the test runner's configuration, these tests fail. To get the tests that interact with the date drive to pass without reimplementing the hooks, I had to temporarily disable these tests. The tests are not irrelevant, but there is a different issue to replace component will receive props.

bumpusfrancus:
Yes, Vanilla date driver mimics moment for all the features used by the app, it does not provide any other part of Moment's API though. In the next version, I would recommend not exposing the date driver to custom components, and just use ISO8601 so that custom components can handle times however they like. That way, the date driver could be made non-configurable. The tests are already not passing. Requiring them to pass makes this issue blocked by #642. I'll look at writing some documentation.

Falkyouall:
is this getting merged sometime soon? I would love to get rid of momentjs

onewaypub:
This right here. We use the library too and it would be great if the problem would be fixed

bumpusfrancus:
> Does it support timezones?

All it does is use native js Date for formatting dates with moment date formats. It relies on the moment formats to maintain backwards compatibility. It can be replaced with a different library in the implementation.

TomHiller-swd:
@bumpusfrancus This works great but I noticed a bug where the mediumLong date format did not convert properly.  It displayed as `L, 01:00`, `L, 02:00`, etc.

lnagel:

This PR introduced a `dateDriver` abstraction layer to replace moment.js with a pluggable date system. The codebase has since taken a different approach: **moment.js has already been fully replaced with dayjs** as a direct dependency — there are zero `moment` imports remaining in the source. All the files this PR modifies (`src/lib/Timeline.js`, `src/lib/columns/Columns.js`, `src/lib/items/Item.js`, `src/lib/utility/calendar.js`) no longer exist as `.js` files — they have been rewritten as TypeScript (`.tsx`/`.ts`). The custom `dateDriver` abstraction (`src/lib/utility/dateDriver.js`, `src/lib/utility/dateDrivers/vanillaDateDriver.js`) was never adopted; instead, dayjs is imported directly throughout the codebase. The test files have also been migrated from Enzyme to @testing-library/react.

The current version is **0.30.0-beta.5** and uses TypeScript, dayjs, Vite (instead of Webpack), and @testing-library/react (instead of Enzyme).

Issue #805, opened 2021-02-25
horizontal scroll isue


StoikovOleh:
**Describe the bug**
horizontal scroll is ignoring visibleTimeStart and visibleTimeEnd, you can scroll timeline even without zoom

**To Reproduce**
in version 26.0.4 everything is ok, problem starting from 26.0.4

you can check how it should be here: https://codesandbox.io/s/timeline-demo-controlled-visible-time-no-scroll-forked-jdw0l
and how it is now here: http://codesandbox.io/s/timeline-demo-controlled-visible-time-no-scroll-forked-h6r6g

xeleoss:
Same problem.

ahwelgemoed:
I had a similar issue with `0.27.0`, upgrading to `0.28.0` did fix it for me

developer-ocansey:
Closing as part of issue triage. This issue was created before 2023 and has had no recent activity. If you're still experiencing this on v0.30+, please open a new issue with a minimal reproduction.

PR #824, opened 2021-06-08
Change the way onWheel event is handled

obadakhalili:
* Possible addressed Issue: [805](https://github.com/namespace-ee/react-calendar-timeline/issues/805), and probably more.
* Overview:
1. Removes unnecessary `onScroll` event in `Timeline.js`.
2. Handles scrolling properly in controlled mode.
3. Synchs header and timeline.

### DIFF
--- a/src/lib/scroll/ScrollElement.js
+++ b/src/lib/scroll/ScrollElement.js
@@ -22,14 +22,6 @@ class ScrollElement extends Component {
     }
   }
 
-  /**
-   * needed to handle scrolling with trackpad
-   */
-  handleScroll = () => {
-    const scrollX = this.scrollComponent.scrollLeft
-    this.props.onScroll(scrollX)
-  }
-
   refHandler = el => {
     this.scrollComponent = el
     this.props.scrollRef(el)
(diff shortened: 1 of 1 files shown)


src/lib/scroll/ScrollElement.js
```diff
@@ -40,13 +32,10 @@ class ScrollElement extends Component {
   
 
   handleWheel = e => {
-    const { traditionalZoom } = this.props
-
-    
+    e.preventDefault()
```
DarioG, line 35: This is preventing the horizontal scroll from working. I think this can be fixed with 

```
handleWheel = e => {
    const isHorizontalScrolling = e.wheelDeltaY === 0

    // zoom in the time dimension
    if (e.ctrlKey || e.metaKey || e.altKey) {
      e.preventDefault()
      const parentPosition = getParentPosition(e.currentTarget)
      const xPosition = e.clientX - parentPosition.x

      const speed = e.ctrlKey ? 10 : e.metaKey ? 3 : 1

      // convert vertical zoom to horiziontal
      this.props.onWheelZoom(speed, xPosition, e.deltaY)
    } else if (e.shiftKey) {
      e.preventDefault()
      // shift+scroll event from a touchpad has deltaY property populated; shift+scroll event from a mouse has deltaX
      this.props.onScroll(this.scrollComponent.scrollLeft + (e.deltaY || e.deltaX))
      // no modifier pressed? we prevented the default event, so scroll or zoom as needed
    } else if (isHorizontalScrolling) {
      e.preventDefault()
      this.props.onScroll(this.scrollComponent.scrollLeft + e.deltaX) 
      }
    }

  }
```
Also, this will cause performance issues, so you should consider adding `window.requestAnimationFrame` when calling `this.props.onScroll(...) `

StoikovOleh: Hi @DarioG, please check the pr: https://github.com/namespace-ee/react-calendar-timeline/pull/834 

fixed this issue it was related to the another method

PR #827, opened 2021-06-23
Fix/showing selected items without canvas verify

RobertDS07:
**Reason of PR**

This PR comes to solution the bug that pops up when you have a button to modify visibleTimeStart and visibleTimeEnd while has a item in drag.

actualy exists a function to provide only that visible items and that function is based on the start_time and end_time of item, but when you are dragging the item and modifying the visibleStartTime and visibleEndTime, the item hasn't the values modifieds, then the calendar don't show it.

Is possible to solution this problem modifying the values when you are dragging the item, but I believe when you are dragging the item the values shouldn't be modify the values and only should modify when drop it and the modify of items state at each event of drag i believe is bad to performance

**Overview of PR**

I only passed a function that already exists via props to the another function that does the verification that I commented above

I don't tested because I can't time to does that, sorry, but I believe is all right

### DIFF
--- a/src/lib/items/Items.js
+++ b/src/lib/items/Items.js
@@ -93,7 +93,13 @@ export default class Items extends Component {
   getVisibleItems(canvasTimeStart, canvasTimeEnd) {
     const { keys, items } = this.props
 
-    return getVisibleItems(items, canvasTimeStart, canvasTimeEnd, keys)
+    return getVisibleItems(
+        items,
+        canvasTimeStart,
+        canvasTimeEnd,
+        keys,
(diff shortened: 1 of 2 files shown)


RobertDS07:
Sorry for my English, I tried 😄

RobertDS07:
The buttons I have commented are these:
image

RobertDS07:
I tested, is not working I solve more latter

Ilaiwi:
@RobertDS07 Please feel free to open an issue with the problem you are facing maybe also a codesandbox example and would be happy to help.

PR #851, opened 2022-02-06
:bug: fix horizontal scroll issue

yuhiSakaguchi:
- related to horizontal scroll issue #805, which says visibleTimeStart, End works to only header. Actually the same is true of onTimeChange.
- This must solve second codesandbox example of #805.


StoikovOleh:
@ys7i it won't, in my project code of  the library responds to your merge request and it just doesn't work properly with trackpad scroll

PR #854, opened 2022-02-18
Add order enforcement in stack process

trasfract:
includes creation of group parameter
shouldStackEnforceOrder

**Issue Number**

In some cases, it is important to enforce the order of items in the same group because they are not of the same type.
In the images below, I have added a few items which are meant to be stacked, but the first two red items are related to the first blue item, the next two red items to the next blue item, and the last red item to the last blue item.

With the current code, what we get does not enforce this order (and there is no way to enforce it)
image

This PR makes sure that an item cannot be above an item which is in horizontal collision.
The resulting screenshot is the following:
image

**Overview of PR**

This PR creates a new group parameter "shouldStackEnforceOrder". 
Clearly, this parameter depends upon isStacked. If isStacked is undefined or false, this new parameter has no effect.
If shouldStackEnforceOrder is defined and true for a group, it uses a slightly modified collision method which ensures order is enforced, as shown on the screenshots above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres (more or less) to [Semantic Versioning](http://semver.o
 * Add unit argument to onZoom and onTimeChange callbacks
 * Add `className` prop to Timeline component to override `react-calendar-timeline` class #682
 * Fix injecting custom vertical line's class names for time periods longer than day
+* Add shouldStackEnforceOrder parameter for groups to keep the consistent order of items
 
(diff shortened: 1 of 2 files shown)


developer-ocansey:
Thank you for this contribution! Unfortunately the codebase has been substantially rewritten since this PR was opened (full TypeScript migration, new test suite, and various architectural changes). This PR is now too outdated to merge cleanly. Please feel free to reimplement this feature against the latest `main` branch — the new codebase should make it easier to build on.

Issue #500, opened 2019-01-24
Scrolling diagonally on a trackpad no longer works. Also the traditionalScroll prop does not work.


SDupZ:
**Describe the bug**
Scrolling diagonally on a trackpad no longer works since version 0.22.0. Additionally the traditionalScroll does prop does not work.

It seems like it was introduced in https://github.com/namespace-ee/react-calendar-timeline/pull/427 in the `src/lib/scroll/ScrollElement.js` file.

**To Reproduce**
You can observe this behaviour in the regular codesandbox: https://codesandbox.io/s/w6xvqzno4w

Gif from library version 0.21.0 (commit 9d0d3fdfd7382d7770af6a8462c11bff3d44acb8)
![jan-24-2019 17-35-04](https://user-images.githubusercontent.com/5068712/51654802-6ea56c80-1ffe-11e9-98aa-1ecc5d987605.gif)

Gif from current library version 0.23.0
![jan-24-2019 17-23-59](https://user-images.githubusercontent.com/5068712/51654457-e377a700-1ffc-11e9-97d1-6bd7c02f2db2.gif)

1. Go to codesanbox
2. Try scroll diagonally on a trackpad

**Expected behavior**
The ability to scroll diagonally when using a trackpad.

**Library Version**
0.22.0 and 0.23.0

**Desktop (please complete the following information):**

* OS: iOS
* Browser: Chrome 71.0.3578.98 and Safari 12.0.2

Ilaiwi:
Thank you for reporting this. I don't think I have the resources to focus on this issue right now so any help is appreciated

ujjwalpandey922:
Hi can I help????

PR #873, opened 2022-09-27
Issue 500 diagonal scroll trackpad

ChrisKurz098:
Issue Number: 500

[Link to issue](https://github.com/namespace-ee/react-calendar-timeline/issues/500)

It seems that "natural scrolling" is a function of the OS or the browser and cannot be changed with JS. By default scrolling is limited to one axis at a time. However, I have added a small bit of code the add this functionality when clicking inside the calendar component. Now the calendar can be clicked and dragged on the X and Y axis.

To do this I turned ```dragStartPosition``` into an object containing  ```x``` and ```y```  scroll values. I then use ```dragStartPosition.y``` to update the windows scroll position.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres (more or less) to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
-
+* Added X and Y dragging on calender component
 
 ## 0.28.0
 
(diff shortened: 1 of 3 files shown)


ChrisKurz098:
there seems to be a property for ```traditionalZoom``` but it doesn't seem to be needed. The traditionalScroll prop mentioned in the issue seems to be undefined

ChrisKurz098:
oops, this was supposed to be a draft!

PR #889, opened 2022-11-28
feat: add prop to remove vertical margin between stacked items

### LINKED ISSUE #888: Row padding
Hello!

I want to display  stacked items next to each other (not vertical margin between them) but have some space between the first and last item from each stack to the horizontal lines of the group row.

image

A possible solution would be to add a prop called `groupRowVerticalPadding: number` to specify the inner padding of each group row. 

This would look something like:
image

franciscohanna92:
closes #888


lnagel:

This PR added a `verticalMarginBetweenStackedItems` boolean prop to control vertical spacing between stacked items. The codebase has since been rewritten from JavaScript to TypeScript (v0.30.0-beta.5), and this feature has been **superseded by a more flexible approach**: the new `itemVerticalGap` prop (a `number`, not a boolean) now provides precise control over the vertical gap between stacked items. It works alongside `itemHeightRatio` and takes precedence when specified. The stacking logic in `src/lib/utility/calendar.tsx` already uses `itemVerticalGap` throughout. Additionally, this PR targeted files that no longer exist in their original form — `src/lib/Timeline.js` is now `src/lib/Timeline.tsx`, tests have been migrated from Enzyme to @testing-library/react, and the `__fixtures__` and `__tests__` files referenced have been restructured.

Issue #838, opened 2021-10-13
onTimeChange called before onItemDrag


atheck:
**Describe the bug**

When dragging an item, `onTimeChange` is called just before `onItemDrag`. This happens only randomly.

In my case this unnecessarily causes reloading of data. When this happens, the moved item gets "recreated" and jumps back to the start position before dragging.

**To Reproduce**

~~~tsx
<Timeline
    groups={groups}
    items={items}
    onItemDrag={() => console.log("onItemDrag")}
    onTimeChange={() => console.log("onTimeChange")}
/>
~~~

Sometimes this will print:

~~~
onTimeChange
onItemDrag
~~~

**Expected behavior**

`onTimeChange` should not be called before dragging an item.

**Library Version**

"react-calendar-timeline": "0.27.0"

**Desktop (please complete the following information):**

* OS: Windows 10
* Browser: Firefox, Chrome
* Version: latest

Ilaiwi:
@atheck this happens because how the library is implemented it's hard to distinguish between a select and drag an item vs dragging the timeline to scroll.
Please feel free to submit a PR with a fix for this issue.

mihir224:
@atheck can I work on this issue?

mihir224:
@atheck can I work on this issue?

pancholipratham7:
Hey I want to contribute to this issue can you please assign this task to me

DanielDerma:
i made a solution: PR: #893

ShabinMuhammedKK:
Add conditions to trigger onTimeChange only when there's a meaningful change in the timeline (i.e., not during a drag)

const handleTimeChange = () => {
  if (!isDragging) {
    console.log("onTimeChange");
  }
};

<Timeline
  groups={groups}
  items={items}
  onItemDrag={() => {
    setIsDragging(true);
    console.log("onItemDrag");
  }}
  onTimeChange={handleTimeChange}
/>

PR #893, opened 2022-12-29
fix: On time change called (no before) on item drag

DanielDerma:
#838

[Link issue.](https://github.com/namespace-ee/react-calendar-timeline/issues/838)

I think that is a misunderstanding with the bug, why it should have been called onTimeChange when I drag i item?

docs:

## onTimeChange(visibleTimeStart, visibleTimeEnd, updateScrollCanvas, unit)
A function that's called when the user tries to **scroll**. Call the passed `updateScrollCanvas(start, end)` with the updated visibleTimeStart and visibleTimeEnd (as unix timestamps in milliseconds) to change the scroll behavior, for example to limit scrolling.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -607,6 +607,9 @@ Plenty of bugfixes, tests and new demos in these 0.14 patch releases.
 * Fix demo for IE11 #44 by @lucidlemon
 * Package a .css file, not a .scss file as previously done. @mariusandra
 
+### fix
+* fix the isInteractingWithItem variable
+
 [0.9.0]: https://github.com/namespace-ee/react-calendar-timeline/compare/v0.8.6...v0.9.0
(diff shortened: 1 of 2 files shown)


Ilaiwi:
@florianbepunkt @MertHaddad can you help verify this PR please

lnagel:

This PR modified `src/lib/Timeline.js` (a JavaScript class component) to include `selectedItem` in the `isInteractingWithItem` check, which controlled whether `onTimeChange` was triggered during item drags. The entire codebase has since been rewritten from JavaScript to TypeScript — the target file is now `src/lib/Timeline.tsx`. Additionally, the scroll and interaction logic has been refactored in v0.30.0-beta.5, and the `isInteractingWithItem` variable in the current code (`!!draggingItem || !!resizingItem`) is used in a different context. The specific behavioral issue from #838 would need to be re-evaluated against the current TypeScript implementation to determine if it still exists.

PR #896, opened 2023-02-02
check for parentNode before using

Nikhil1489:
**Issue Number**

Link to issue this PR addresses.  If no issue exists, please provide reasoning for PR.

**Overview of PR**

### DIFF
--- a/src/lib/utility/dom-helpers.js
+++ b/src/lib/utility/dom-helpers.js
@@ -25,8 +25,8 @@ export function getParentPosition(element) {
 }
 
 export function getSumScroll(node) {
-  if (node === document.body) {
-    return {scrollLeft: 0, scrollTop: 0}
+  if (node === document.body || !node.parentNode) {
+    return { scrollLeft: 0, scrollTop: 0 }
   } else {
     const parent = getSumScroll(node.parentNode)
     return ({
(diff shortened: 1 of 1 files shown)


lnagel:
@developer-ocansey — Flagging this for your review.

**Recommendation: The bug still exists — apply as a quick fix to the current codebase**

### The bug
`getSumScroll` in `dom-helpers.ts` recursively traverses `node.parentNode` but doesn't guard against it being null. If a node is detached from the DOM or inside a shadow DOM, `parentNode` can be null, causing a crash on the next recursion.

The current code (line ~33 of `dom-helpers.ts`):
```typescript
export function getSumScroll(node: HTMLElement): { scrollLeft: number; scrollTop: number } {
  if (node === document.body) {
    return { scrollLeft: 0, scrollTop: 0 }
  } else {
    const parent = getSumScroll(node.parentNode as HTMLElement)
    // ...
  }
}
```

Note that the sibling function `getSumOffset` already has an analogous guard (`!node.offsetParent`) at line ~50, so adding `!node.parentNode` to `getSumScroll` would be consistent.

### Why it can't merge directly
The PR targets the old `dom-helpers.js` — the fix needs to be re-applied to `dom-helpers.ts`. It's a 1-line guard addition.

I'd suggest just applying the fix directly to `dom-helpers.ts` rather than trying to rebase this PR.

PR #913, opened 2023-08-23
FIx redraw mechanism to be based on buffer and not viewport

LudvigHz:
Also updates node-sass to latest version to support newer node versions.

The README states that the canvas should redraw when there is 1/2 of the buffer left off-screen on any side of the viewport. However, the code calculates the redraw based on 1/2 of the _viewport_ and not the _buffer_. This PR fixes this. This means that for buffers larger than the default, we re-render way less often, providing for better performance with large timelines. As it is now, for many tasks (~1000) the canvas continuously re-renders when scrolling, and the rendering cannot keep up with scrolling, forcing the items to appear as they jump around a lot.

I don't know if this repo is actively maintained, but we will use our [own fork](https://github.com/TaskCtrl-AS/react-calendar-timeline/) ([NPM](https://www.npmjs.com/package/@taskctrl/react-calendar-timeline)) as long as this PR is open and the repo appears unmaintained.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,8 +5,10 @@
   "main": "lib/index.js",
   "scripts": {
     "build": "npm run build:lib",
-    "build:demo": "echo '!!! Building Demo' && cross-env NODE_ENV=production webpack --progress",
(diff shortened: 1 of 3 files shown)


LudvigHz:
This is obviously not maintained, so I'm going ahead and closing this :)

Ilaiwi:
Hi @LudvigHz, we released a new version with updated dependencies. Please feel free to test out the new version and reopen this PR.

LudvigHz:
Thanks @Ilaiwi. I rebased the changes on master and opened a new PR: #936

PR #936, opened 2024-11-17
Use buffer prop to calculate when to redraw canvas

LudvigHz:
This is a rebased version of #913

**Overview of PR**

The README states that the canvas should redraw when there is 1/2 of the buffer left off-screen on
any side of the viewport. However, the code calculates the redraw based on 1/2 of the viewport and
not the buffer. This PR fixes this. This means that for buffers larger than the default, we
re-render way less often, providing for better performance with large timelines. As it is now, for
many tasks (~1000) the canvas continuously re-renders when scrolling, and the rendering cannot keep
up with scrolling, forcing the items to appear as they jump around a lot.


Ilaiwi:
@LudvigHz thank you, I am curious, did you get a chance to test out the new beta release with your app? have you seen any issues?

Issue #940, opened 2024-11-24
Add option do drop new tasks to the timeline from an external component


Remco4EF:
**Is your feature request related to a problem? Please describe.**
We want to be able to drop tasks from an external list of items to the timeline. And plan it afterwards.

**Describe the solution you'd like**
With interactjs it is not to difficult to receive the dropped object. Only the Timeline should tell me which x and y components translate to which group and time.

The timeline component should provide a callback that gives these values back

Issue #939, opened 2024-11-24
Externalize stylesheet instead of bundling with libInjectCss vite plugin


Remco4EF:
**Is your feature request related to a problem? Please describe.**
With the 0.30 rewrite somewhere we made the error of bundling the stylesheet in the package.
Great as it has default styling attached to it, but harder to override.

Also, nextjs pages router complains about this.

**Describe the solution you'd like**
- add the default stylesheet to the package
- add the sass file to the package

So you can choose which file to use.
`import Timeline from 'react-calendar-timeline';`
and
`import 'react-calendar-timeline/dist/Timeline.scss';`
or
`import 'react-calendar-timeline/style.css';`

**Additional context**
While working with the package, some type definitions got enhanced.

PR #941, opened 2024-11-24
Add option do drop new tasks to the timeline from an external component

Remco4EF:
This pull request proposes a solution for #940 

We suggest a table (to the right) containing draggable elements. These can be dropped onto the timeline canvas.
Since rct already uses interactjs, why not use that to reach this goal?

image

**Overview of PR**

All logic on what to drop at which moment is allowed, is up to the user via the interact configuration.
The library needs to expose some callbacks in order to be able to calculate where the new item is dropped.

Also #939 is fixed with this PR, so no bundled stylesheet anymore. 
And basic demo with nextjs 15 is added.


Tvechko:
@Ilaiwi Is there an ETA for when this PR gets merged/next beta is released? Would love to have the stylesheet fix for my nextjs project.

Ilaiwi (approved):

